PR #2, opened 2019-08-23
Commands for working with arrays

jgallag88:


sdb/commands/array.py
```diff
@@ -0,0 +1,40 @@
...
+import drgn
+import sdb
+
+class Array(sdb.Command):
+    # pylint: disable=too-few-public-methods
+
+    names = ["array"]
+
+    def _init_argparse(self, parser: argparse.ArgumentParser) -> None:
+        # TODO allow 'count' to be omitted if input type is proper array
```
prakashsurya: Does this comment still apply? i.e. do we expect this command to actually work if the input type is not a proper array? Specifically, I wonder if we can use the index operator (on line 38) if the input type is not a "proper array"..?

I could be wrong, but I feel like we should remove this comment, but leave the rest of the implementation as-is. What do you think?

jgallag88: This comment still applies. The index operator will work with any pointer type (except `void*`), not just arrays:
```
> ... | member svc_export_cache hash_table
*(struct hlist_head *)0xffff9b643ffe4000 = {
	.first = (struct hlist_node *)0x0,
}
> ... | member svc_export_cache hash_table | array 4
(struct hlist_head){
	.first = (struct hlist_node *)0x0,
}
(struct hlist_head){
	.first = (struct hlist_node *)0x0,
}
(struct hlist_head){
	.first = (struct hlist_node *)0x0,
}
(struct hlist_head){
	.first = (struct hlist_node *)0x0,
}
```

jgallag88:
> Also, can you add some tests for these new commands

Sure, I'll look into adding some tests.

sdb/commands/array.py
prakashsurya: We'll have to either convert the comment to something that doesn't have "TODO" in it, or suppress the pylint error:
```
sdb/commands/array.py:31:2: W0511: TODO allow 'count' to be omitted if input type is proper array (fixme)
```

IMO, if we want to create a version of this where "count" doesn't have to be provided, I think that might be better served by a new command that inherits from `sdb.Walker`; and leaving this to some future change.

But I'm OK leaving this as-is, with the comment, if that's what you'd prefer.

sdb/commands/array.py
```diff
@@ -0,0 +1,40 @@
...
+        parser.add_argument("count", type=int, metavar="<count>")
```
prakashsurya: I'm OK leaving this for a future change (i.e. feel free to drop this), but it might be a useful addition to enable a range to be supplied here, rather than simply the first N entries.. and perhaps use the python syntax for the `[]` operator.. e.g. `:N`, `N:`, `:-1`, etc.

prakashsurya: This might actually be better to be implemented by the `index` command.. and then this `array` command might be unnecessary.

I don't expect you to make any changes, but just writing down my thoughts as I'm looking over this.

sdb/commands/array.py
sdimitro: I think implementing this TODO should be straightforward, playing around with `drgn` you can find that length easily as arrays have a `length` field:
```
>>> prog.type('struct spa').members[0]
(array_type(length=256, type=int_type(name='char', size=1, is_signed=True)), 'spa_name', 0, 0)
```

As for the walker idea that Prakash brought up here are some of my thoughts:
[1] We can have this `array` command take the following optional fields `<entry_size>` and `<count>`. If none of the fields are provided, then we look whether this is an array type with the length and walk it. If none, of the fields are provided and we are passed a `type_t` double, triple, etc.. pointer, we can get the `entry_size` from the machine architecture (e.g. 64-bit means 8 bytes) but we need to error out because we can't tell the count, thus `count` should be supplied in these situations. If a `type_t` single pointer is provided, we can get the entry size by the `type_t` but we still need to error out if `count` is not provided. If a `void *` is passed, we error out in both `count` and `entry_size`. If an array type is passed, we don't need `entry_size` or `count` as we can get those from the type itself (one exception that we may need to experiment is VLAs though!). Still these can be overwritten if a `count` or an `entry size` argument is passed.

[2] Once `array` is in place, then we can create a walker on top of it

jgallag88:
An alternative implementation of this command was added in https://github.com/delphix/sdb/pull/127

PR #16, opened 2019-10-01
Improve unit test mocks by introducing more complex structures

sdimitro:
= Motivation

This change is part of a bigger change that I'm planning to
upstream soon - the new and improved `member` command. In
order to test the command more thoroughly I needed structures
in our mocks that are a bit more complex than the existing
one. As the code change grew bigger, I decided to split it
into the testing infrastructure part and the actual command
part.

= Patch

This patch introduces adds a new array field in the test_struct
structure, and then introduces another struct that contains
test_struct pointers and an embedded struct. All of these cases
will be used by the tests added together with the new `member`
command.

= Testing

Besides ensuring that current tests don't fail I manually
checked the values of all the new mocks from the python
REPL like this:
```
$ python3 -i tests/__init__.py

>>> print(MOCK_PROGRAM['global_struct'])
(struct test_struct){
        .ts_int = (int)1,
        .ts_voidp = (void *)0xffffffffc0000000,
        .ts_array = (int [2]){ 252645135, -559038737, },
}

>>> print(hex(MOCK_PROGRAM['global_int'].address_of_().value_()))
0xffffffffc0000000
>>> print(hex(MOCK_PROGRAM['global_struct'].ts_array[0]))
0xf0f0f0f
>>> print(hex(drgn.cast('unsigned int', MOCK_PROGRAM['global_struct'].ts_array[1])))
0xdeadbeef

>>> print(MOCK_PROGRAM['global_cstruct'])
(struct complex_struct){
        .cs_structp = (struct test_struct *)0xffffffffc0a8aee0,
        .cs_struct = (struct test_struct){
                .ts_int = (int)10,
                .ts_voidp = (void *)0xffffffffd0000000,
                .ts_array = (int [2]){ -2147483648, 2147483647, },
        },
        .cs_structp_null = (struct test_struct *)0x0,
}

>>> print(hex(MOCK_PROGRAM['global_struct'].address_of_().value_()))
0xffffffffc0a8aee0
>>> print(hex(MOCK_PROGRAM['global_cstruct'].address_of_().value_()))
0xffffffffd0000000
>>> print(hex(MOCK_PROGRAM['global_cstruct'].cs_struct.ts_array[0]))
-0x80000000
>>> print(hex(MOCK_PROGRAM['global_cstruct'].cs_struct.ts_array[1]))
0x7fffffff
>>> print(hex(MOCK_PROGRAM['global_cstruct'].cs_structp_null))
0x0
```

### DIFF
--- a/tests/commands/test_member.py
+++ b/tests/commands/test_member.py
@@ -60,7 +60,7 @@ def test_member_not_found():
     assert "'struct test_struct' has no member 'bogus'" in str(err.value)
 
 
-def test_one_member():
+def test_first_member():
     line = 'addr global_struct | member ts_int'
     objs = []
 
@@ -70,3 +70,17 @@ def test_one_member():
     assert ret[0] == drgn.Object(MOCK_PROGRAM,
                                  MOCK_PROGRAM.type('int'),
                                  value=1)
+
+
+def test_second_member():
+    line = 'addr global_struct | member ts_voidp'
+    objs = []
+
+    ret = invoke(MOCK_PROGRAM, objs, line)
+
+    #
+    # We expect global_struct.ts_voidp to point to
+    # global_int (see comment on fake_mappings).
+    #
+    assert len(ret) == 1
+    assert ret[0] == MOCK_PROGRAM['global_int'].address_of_()
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -46,12 +46,18 @@ def create_struct_type(name: str, member_names: List[str],
     for member_name, type_ in zip(member_names, member_types):
         member_tuple = (type_, member_name, bit_offset, 0)
         member_list.append(member_tuple)
-        bit_offset += 8 * struct_size
-        struct_size += type_.size
+        if type_.kind == drgn.TypeKind.ARRAY:
+            bit_offset += 8 * type_.length * type_.type.size
+            struct_size += type_.length * type_.type.size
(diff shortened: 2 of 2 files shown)


sdimitro:
= Notes on Future

Hopefully, this will be the last commit in the mock infrastructure that
has to add symbols and types like that. The current way is cumbersome,
tedious, and error-prone due to bad readability. Some ideas for future
versions of this test mocks are listed below:
[1] Save all the drgn state from a sample drgn session with a crash dump
    in Python's serializable format and reload that state in the testing
    suite. This should allow us to tests the zfs commands too.
[2] Have filtered crash dumps for each test group/unit that we load on
   demand.

tests/commands/test_member.py
```diff
@@ -70,3 +70,13 @@ def test_one_member():
...
+
+
+def test_second_member():
+    line = 'addr global_struct | member ts_voidp'
+    objs = []
+
+    ret = invoke(MOCK_PROGRAM, objs, line)
+
+    assert len(ret) == 1
+    assert ret[0] == MOCK_PROGRAM['global_int'].address_of_()
```
prakashsurya: I don't feel strongly, but it might be worth a quick comment/hint to the reader, that "ts_voidp" points to the "global_int" variable. It took me a second to understand this line, but it makes sense now that I realize the value of the pointer.

PR #20, opened 2019-10-03
Introduce slabs command

sdimitro:
= Motivation

`proc/slabinfo` is a useful tool for debugging out-of-memory
and fragmentation issues for the kernel. Unfortunately, it is
not applicable to crash dumps and not very user friendly overall
(e.g. you could figure out the memory usage of each cache and
its utilization but you have to do the math on your own).

Besides the above, the specific use-case within Delphix is a
replacement for `::kmastat` that we had in MDB in the illumos
version of the product. The `slabs` command introduced in this
commit is not a full replacement as it gives you only the
memory managed by the slab allocator (e.g. no memory zone, no
vmem info, etc..) and we still need to implement it for linux.
The idea here though, is to implement all the various memory
commands separately and then have `kmastat` finally consume
all the output from these commands.

= Patch

The main portion of this patch is the introduction of the slabs
command. By default the command iterates over the root caches
and prints the following statistics in human-readable form
sorted by memory used:
```
> slabs
name                        entry_size  objs mem_used util
--------------------------- ---------- ----- -------- ----
zfs_znode_cache                   1104 35224   37.1MB   99
dnode_t                            912 38284   33.3MB   99
zio_buf_14336                    16384  1916   29.9MB   99
kmalloc-512                        512 47824   23.4MB   93
dentry                             192 98112   18.0MB   97
...
```

All the properties can be seen in the help message and the user
may specify the properties they are looking for with the `-o`
option (and also sort them by a specific field using `-s`):
```
> slabs -h
usage: slabs [-h] [-H] [-o FIELDS] [-p] [-r] [-s FIELD] [-v]

optional arguments:
  -h, --help       show this help message and exit
  -H               do not display headers and separate fields by a single tab
                   (scripted mode)
  -o FIELDS        comma-separated list of fields to display
  -p               display numbers in parseable (exact) values
  -r, --recursive  recurse down children caches and print statistics
  -s FIELD         sort rows by FIELD
  -v               Print all statistics

FIELDS := address, name, object_size, entry_size, entries_per_slab, slabs,
mem_used, objs, active_objs, inactive_objs, util

If -o is not specified the default fields used are name, entry_size, objs,
mem_used, util.

If the -s option is not specified and the command's output is not piped anywhere
then we sort by the following fields in order: mem_used, address, name. If none
of those exists in the field-set we sort by the first field specified in the
set.

> slabs -o "name,object_size,slabs" -s object_size
name                        object_size slabs
--------------------------- ----------- -----
zio_buf_14336                     16384   957
zio_buf_10240                     12288    13
kmalloc-8192                       8192   527
dma-kmalloc-8192                   8192     0
net_namespace                      6912     2
...
```

Besides a PrettyPrinter, the command is also a Locator which
means that the user can do things like this:
```
> slabs | filter obj.name == "dma-kmalloc-192" | slabs
name            entry_size objs mem_used util
--------------- ---------- ---- -------- ----
dma-kmalloc-192        192    0     0.0B    0
```

The user can also use the `-r` option to traverse the children
of the root-caches recursively. Unfortunately, our `Table`
implementation doesn't support grouping entries together so
in our case we can't really group children caches and sort
them as one entry under their respective parent cache. This
has been left for future work and the current limited workaround
is to ensure that the entries are sorted by name:
```
> slabs -r -s name
name                                                       entry_size  objs mem_used util
---------------------------------------------------------- ---------- ----- -------- ----
AF_VSOCK                                                         1152    14   15.8KB  100
Acpi-Namespace                                                     40 46206    1.8MB   99
Acpi-Operand                                                       72  5488  385.9KB  100
Acpi-ParseExt                                                     104   156   15.8KB  100
Acpi-State                                                         80  5559  434.3KB   67
RAW                                                               960    34   31.9KB  100
TCP                                                              2048    96  192.0KB  100
anon_vma                                                           88 20654    1.7MB  100
anon_vma(107:systemd-remount-fs.service)                           88     0     0.0B    0
anon_vma(1086:systemd-tmpfiles-clean.service)                      88     0     0.0B    0
anon_vma(1118:motd-news.service)                                   88     0     0.0B    0
anon_vma(1158:motd-news.service)                                   88     0     0.0B    0
...
```

A couple more options mimicking Illumos conventions are the `-p`
and `-H` options that output numbers to raw form and skip printing
the headers respectively. This is generally useful for scripts or
output that is to be processed later:
```
> slabs -H -p
zfs_znode_cache	1104	35224	38887296	100
dnode_t	912	38318	34946016	99
zio_buf_14336	16384	1932	31653888	99
kmalloc-512	512	47824	24485888	93
dentry	192	98112	18837504	97
```

As part of the command's implementation, some generic infrastrure
was introduced in the repo. The `Table` class is the biggest
part of these changes, that is expected to be re-used by many
other commands. It was originally written by Matt Ahrens and I
had to edit it a bit to be Python3, yapf, and pylint compliant.
I also added the `formatters` field to support user-supplied
pretty-printers.

= Testing/Verification

I verified the values of all their fields in the following ways:
[1] For the fields that are also visible in /proc/slabinfo, I
    ensured that the values matched.
[2] For anything else I did the math and cross-referenced all
    the values ensuring that they are consistent with each other.

Besides the manual testing done shown in the output above I also
ensured the following error cases:

* ask for invalid field name
```
> slabs -o bogus
sdb: slabs: 'bogus' is not a valid field
```

* sort by invalid field name
```
> slabs -s bogus
sdb: slabs: invalid input: 'bogus' is not in field set (name, entry_size, objs, mem_used, util)
```

* attempt to sort by field that is not in the requested
field set
```
> slabs -o "name,address" -s objs
sdb: slabs: invalid input: 'objs' is not in field set (name, address)
```

= Other Notes

[1] The command currently only works with SLUB from kernel version
    4.15+. It will probably crash and burn with other allocators.
[2] Since we still don't do proper lexing in the SDB repl the `-o`
    option only works for multiple properties when these are
    quoted together as a string and there are not spaces between
    the fields/commas. To give an example:
    `> slabs -o name,util` <-- This doesn't work (yet).
    `> slabs -o "name, util"` <-- This doesn't work.
    `> slabs -o "name,util"` <-- This is the current workaround.

### DIFF
--- a/sdb/commands/internal/table.py
+++ b/sdb/commands/internal/table.py
@@ -1,0 +1,91 @@
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pylint: disable=missing-docstring
+
+from typing import Any, Callable, Dict, List, Optional, Set
+
+
+class Table:
+    """
+    Generic Table implementation for pretty-printing
+    data in columns.
+    """
+
+    __slots__ = "fields", "rjustfields", "formatters", "maxfieldlen", "lines"
+
+    def __init__(self,
+                 fields: List[str],
+                 rjustfields: Optional[Set[str]] = None,
+                 formatters: Optional[Dict[str, Callable[[Any], str]]] = None
+                ) -> None:
+        self.fields = fields
+
+        if rjustfields is None:
+            self.rjustfields = ()
+        else:
+            self.rjustfields = rjustfields
+
+        self.formatters = {
+            field: (formatters[field] if field in formatters.keys() else str)
+            for field in fields
+        }
+        self.maxfieldlen = dict.fromkeys(fields, 0)
+        self.lines = []
+
+    def add_row(self, sortkey: Any, values: Dict[str, Any]) -> None:
+        row_values = []
+        for field in self.fields:
+            val = self.formatters[field](values[field])
+            row_values.append(val)
+            self.maxfieldlen[field] = max(self.maxfieldlen[field], len(val))
+        self.lines.append((sortkey, row_values))
+
+    def print_(self,
+               print_headers: Optional[bool] = True,
+               reverse_sort: Optional[bool] = False) -> None:
+        delimeter = "\t"
+        if print_headers:
+            delimeter = " "
+            headers, separators = [], []
+            for field in self.fields:
+                self.maxfieldlen[field] = max(self.maxfieldlen[field],
+                                              len(field))
+                if field in self.rjustfields:
+                    headers.append(f"{field:>{self.maxfieldlen[field]}}")
+                else:
+                    headers.append(f"{field:<{self.maxfieldlen[field]}}")
+                separators.append('-' * self.maxfieldlen[field])
+            print(delimeter.join(headers))
+            print(delimeter.join(separators))
+
+        self.lines.sort(reverse=reverse_sort)
+        for _, row_values in self.lines:
+            if not print_headers:
+                print(delimeter.join(row_values))
+                continue
+
+            line_fields = []
+            for fid in range(len(self.fields)):
+                if self.fields[fid] in self.rjustfields:
+                    line_fields.append(
+                        f"{row_values[fid]:>{self.maxfieldlen[self.fields[fid]]}}"
+                    )
+                else:
+                    line_fields.append(
+                        f"{row_values[fid]:<{self.maxfieldlen[self.fields[fid]]}}"
+                    )
+            print(delimeter.join(line_fields))
--- a/sdb/commands/linux/internal/slub_helpers.py
+++ b/sdb/commands/linux/internal/slub_helpers.py
@@ -1,0 +1,118 @@
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pylint: disable=missing-docstring
+
+from typing import Iterable
+
+import drgn
+from drgn.helpers.linux.list import list_for_each_entry
+
+
+def is_root_cache(obj: drgn.Object) -> bool:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    return obj.memcg_params.root_cache.value_() == 0x0
+
+
+def list_for_each_root_cache(prog: drgn.Program) -> Iterable[drgn.Object]:
+    yield from list_for_each_entry("struct kmem_cache",
+                                   prog["slab_root_caches"].address_of_(),
+                                   "memcg_params.__root_caches_node")
+
+
+def list_for_each_child_cache(root_cache: drgn.Object) -> Iterable[drgn.Object]:
+    assert root_cache.type_.type_name() == 'struct kmem_cache *'
+    yield from list_for_each_entry(
+        "struct kmem_cache", root_cache.memcg_params.children.address_of_(),
+        "memcg_params.children_node")
+
+
+def nr_slabs(obj: drgn.Object) -> int:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    nslabs = obj.node[0].nr_slabs.counter.value_()
+    if not is_root_cache(obj):
+        return nslabs
+    for child in list_for_each_entry("struct kmem_cache",
+                                     obj.memcg_params.children.address_of_(),
+                                     "memcg_params.children_node"):
+        nslabs += nr_slabs(child)
+    return nslabs
+
+
+def entries_per_slab(obj: drgn.Object) -> int:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    return obj.oo.x.value_() & 0xffff
+
+
+def entry_size(obj: drgn.Object) -> int:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    return obj.size.value_()
+
+
+def object_size(obj: drgn.Object) -> int:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    return obj.object_size.value_()
+
+
+def memused(obj: drgn.Object) -> int:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    nslabs = nr_slabs(obj)
+    epslab = entries_per_slab(obj)
+    esize = entry_size(obj)
+    return nslabs * epslab * esize
+
+
+def objs(obj: drgn.Object) -> int:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    count = obj.node[0].total_objects.counter.value_()
+    if not is_root_cache(obj):
+        return count
+    for child in list_for_each_entry("struct kmem_cache",
+                                     obj.memcg_params.children.address_of_(),
+                                     "memcg_params.children_node"):
+        count += objs(child)
+    return count
+
+
+def inactive_objs(obj: drgn.Object) -> int:
+    assert obj.type_.type_name() == 'struct kmem_cache *'
+    node = obj.node[0].partial
+    free = 0
+    for page in list_for_each_entry("struct page", node.address_of_(), "lru"):
+        free += page.objects.value_() - page.inuse.value_()
+
+    if not is_root_cache(obj):
+        return free
+
+    for child in list_for_each_entry("struct kmem_cache",
+                                     obj.memcg_params.children.address_of_(),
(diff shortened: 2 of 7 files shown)


sdimitro:
cc @ahrens

sdb/commands/internal/table.py
```diff
@@ -0,0 +1,93 @@
...
+                ) -> None:
+        self.fields = fields
+
+        if rjustfields is None:
+            self.rjustfields = ()
+        else:
+            self.rjustfields = rjustfields
+
+        self.formatters = {field: str for field in fields}
+        if not formatters is None:
```
ahrens: is it more idiomatic to do `if formatters is not None:`?

sdb/commands/internal/table.py
```diff
@@ -0,0 +1,93 @@
...
+            for field in formatters.keys():
+                if field in self.fields:
+                    self.formatters[field] = formatters[field]
```
ahrens: I wonder if it would be easier to understand this as:

```
self.formatters = {}
for field in fields:
  if field in formatters.keys():
    self.formatters[field] = formatters[field]
  else
    self.formatters[field] = str
```

Or maybe I'm just not thinking "pythonic" enough.  There's probably a dictionary comprehension that's even more compact.

sdb/commands/internal/table.py
```diff
@@ -0,0 +1,93 @@
...
+        row_values = []
+        for field in self.fields:
+            val = self.formatters[field](values[field])
+            row_values.append(val)
+            self.maxfieldlen[field] = max(self.maxfieldlen[field], len(val))
+        self.lines.append((sortkey, row_values))
+
+    def print_(self,
+               print_headers: Optional[bool] = True,
+               reverse_sort: Optional[bool] = True) -> None:
```
ahrens: reverse_sort defaults to true?  That seems like a strange semantic to me.

ahrens:
Can we make the help message document the defaults (e.g. for `-o` and `-s`)?

How do you feel about "properties" vs "fields" in the documentation? (`-o PROPERTIES` vs `-o FIELDS`)?  "field" is what we use in the zfs command documentation.

sdb/commands/linux/slabs.py
```diff
@@ -0,0 +1,141 @@
...
+                            action='store_true',
+                            help='display numbers in parseable (exact) values')
+        parser.add_argument(
+            '-r',
+            '--recursive',
+            action='store_true',
+            help='recurse down children caches and print statistics')
+        parser.add_argument('-s',
+                            metavar="PROPERTY",
+                            help='sort rows by PROPERTY')
```
prakashsurya: I'm not sure what's expected in this case, but I was experimenting with this command and noticed the following inconsistency:
```
> slabs -s mem_used ! head -n3
name                        entry_size   objs mem_used util
--------------------------- ---------- ------ -------- ----
kmalloc-512                        512 239296  116.8MB   51

> slabs -s mem_used | head 1 | pp
name             entry_size objs mem_used util
---------------- ---------- ---- -------- ----
t10_pr_reg_cache        704    0     0.0B    0
```

It seems as though sorting only works when pretty printing, and not what the objects are piped into another SDB command. Is this a bug, or working as designed?

sdb/commands/linux/slabs.py
```diff
@@ -0,0 +1,141 @@
...
+        parser.add_argument('-v',
+                            action='store_true',
+                            help='Print all statistics')
+        parser.epilog = "PROPERTIES := {:s}".format(','.join(
+            Slabs.PROPERTIES.keys()))
```
prakashsurya: Not a big deal, but where the newline is placed here is a little awkward:
```
PROPERTIES := address,name,object_size,entry_size,entries_per_slab,slabs,mem_u
sed,objs,active_objs,inactive_objs,util
```
i.e. `mem_used` is split across multiple lines.

sdb/commands/linux/slabs.py
sdimitro: Hey Prakash! This is a very interesting point that you brought up and it definitely has to do with how things are designed (specifically the Locator and PrettyPrinter hybrids bundled together with the nature of this command).

Since the command is a `Locator`+`PrettyPrinter` hybrid, it's functionality is currently structured in 2 parts:
[1] The `Locator` part which iterates through the slab caches in the order that we find them in the kernel
[2] The `PrettyPrinter` part which is drains the input that it gets from [1] (or whatever it is passed through the pipeline) in the `for obj in objs:` loop in the beginning of its method, and once that is done it does the sorting by itself and outputs everything.

With this in mind what you have on your example below makes sense:
```
> slabs -s mem_used | head 1 | pp
name             entry_size objs mem_used util
---------------- ---------- ---- -------- ----
t10_pr_reg_cache        704    0     0.0B    0
```
`sdb` figures out that the `slabs` command is not the last one in the pipeline and executes only the `Locator` functionality (part [1] above). The `-s` option is basically discarded as we never get to the pretty-printer part.

I wouldn't say that's a bug but more of a design trade-off. By not sorting in the `Locator` itself we "hold" one `drgn.Object` in memory (a slab cache) that we `yield` along the pipeline to each invocation making each yield `O(1)` in runtime. If we were to sort in the `Locator` and then `yield` each object we'll be holding all the `drgn.Object`s for all slab caches in memory until the `Locator` is fully drained. This would also make the runtime of the first invocation `O(nlogn)` but subsequent ones would still be `O(1)`.

If that trade-off is worth it for `slabs` I'm not sure. I do agree that `-s` being discarded may be misleading. At the same time I'm thinking that there could definitely be other commands in the future where memory could be a problem (e.g. a `pages` `Locator` that walks through all the physical memory pages in the system). And then the trade-off of consistent user experience comes in - do we make things like `-s` discarded for all commands to be consistent or we let each command do its own thing? Maybe the latter is not bad. At the same time I can see us keeping the discarding of options like `-s` consistent and maybe introducing a `sort` command where we leave it to the user to sort the objects in the pipeline in a more explicit way ( this would be similar to how we implemented `filter` instead of adding an option to each command that we want to filter the output from).

I'll  created an issue about this where we can discuss this separately. Definitely a great point.

prakashsurya: That all makes sense from an implementation perspective; and that's what I assumed was the culprit. I still think it's awkward from a UX perspective. With that said, I'm OK going with this as-is, and perhaps revisiting this in the future.

Creating some type of `sort` command, similar to `filter` seems reasonable to me, but then it raises the question of how do we do pretty printing of the sorted information.. e.g. would we still support `slabs -s`, or would we require `slabs | sort | pp` ?

We can continue this discussion in another issue, and/or at a later time. I'm OK moving forward with what you have here.

ahrens: IMO, if I request sorting by passing `-s ...`, the output should be sorted.  The alternative would be for the command to fail, with an error message like "I don't believe you really want to wait for the output to be sorted".  Yes, sorting uses CPU and memory, but if the user asks for it, we should do it.  Folks using a kernel debugger are typically sophisticated enough to figure out what's going on.  If the user gets tired of waiting they can hit ^c and then run it without the `-s`.

The default behavior (when `-s` is not specified) is debatable.  Perhaps it could be unsorted when used in the pipeline but sorted by `mem_used` when pretty-printing.

sdimitro: I opened https://github.com/delphix/sdb/issues/40

sdimitro:
@prakashsurya @ahrens - can you take a second look?

I've updated the following things:
[1] s/properties/fields/
[2] Help message includes defaults for `-o` and `-s`
[3] All `if not x in y` have been converted to `if x not in y`
[4] `reverse_sort` default parameter is `False`
[5] `slabs -s mem_used | head 1 | pp` should work now as discussed
[6] text wrapping at ~80 characters in helper epilog

sdb/commands/internal/util.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def sort_objects_by(objs: Iterable[drgn.Object],
+                    field_getter: Callable[[drgn.Object], Any],
+                    reverse_sort: Optional[bool] = False
+                   ) -> Iterable[drgn.Object]:
+    obj_list = [(field_getter(obj), obj) for obj in objs]
+    obj_list.sort(reverse=reverse_sort)
+    for _, obj in obj_list:
+        yield obj
```
prakashsurya: I feel like this function is unnecessary, as it's just re-implementing the `sorted` builtin ([link](https://docs.python.org/3/library/functions.html#sorted).

Couldn't consumers more simply do this:
```
            yield from sorted(
                self.__no_input_iterator(),
                key=Slabs.FIELDS[self.args.s],
                reverse=(
                    self.args.s not in Slabs.DEFAULT_INCREASING_ORDER_FIELDS))
```
?

sdimitro: Definitely! Applying the change now

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -0,0 +1,108 @@
...
+# pylint: disable=missing-docstring
+
+from typing import Iterable
+
+import drgn
+from drgn.helpers.linux.list import list_for_each_entry
+
+
+def is_root_cache(obj: drgn.Object) -> bool:
+    return obj.memcg_params.root_cache.value_() == 0x0
```
ahrens: when do we need to use `.value()`?  I think the drgn.Object implements `==` and you might be able to do `obj.memcg_params.root_cache == 0x0`?

ahrens: is there a canonical way to check or assert that the drgn.Object is of the expected C type?  Or do we just rely on "duck typing"?  If nothing else, we could declare the expected C type via a comment.

sdimitro: Added asserts to check the name of the C type for those functions is what we expect:
```
+    assert obj.type_.type_name() == 'struct kmem_cache *'
```

sdimitro: Per our conversation, this is unfortunately not possible with the `drgn` API currently:
```
>>> a.memcg_params.root_cache
(struct kmem_cache *)0x0
>>> a.memcg_params.root_cache + 1
(struct kmem_cache *)0x2158
>>> a.memcg_params.root_cache == 0x0
  File "<console>", line 1, in <module>
TypeError: invalid operands to comparison ('struct kmem_cache *' and 'int')
```

PR #48, opened 2019-10-25
Backport master to 6.0/release

sdimitro:


prakashsurya:
How did you pull in these commits? with rebase? or a merge?

sdimitro:
I just checked out master and pushed to a remote branch in my repo called `backports`. Then I opened a pull request targeting `delphix:6.0/release` instead of `master` and Github picked up all the commits that differ. Does this make sense? Do we want to do this in a different way?

prakashsurya:
Cool. Until we have different history for the two branches, I think that's perfect.

PR #50, opened 2019-10-25
Rename existing slub cache walkers

sdimitro:
= Motivation

Make the name shorter - there is no needed to expose
that they're implemented by lists (which could also
theoretically change in the future).

= Testing

Ensured that `slabs` and `slabs -r` runs as before.


sdimitro:
rebased and fixed

PR #54, opened 2019-11-04
Leverage Github Actions for automated testing 

prakashsurya:
This change adds the necessary files and logic to take advantage of
GitHub Actions for running our automated tests. Currently we rely on
TravisCI to do this, but we'd like to transition over to GitHub Actions,
for the following reasons:

  1. The Actions UI is more integrated with GitHub PRs
  2. Best practices at Delphix suggests using Actions

While this change is adding support for using GitHub Actions, it's not
currently disabiling the use of TravisCI. Since I don't have much
experience with Actions, I'd like to use both Actions and TravisCI
simultaneously until we have sufficient confidence in Actions; at which
point we can disable the TravisCI tests, and rely solely on Actions.

### DIFF
--- a/.github/scripts/install-drgn.sh
+++ b/.github/scripts/install-drgn.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -eux
+
+#
+# These are build requirements of "drgn"; if we don't install these, the
+# build/install of "drgn" will fail below.
+#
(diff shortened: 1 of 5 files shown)


prakashsurya:
@sdimitro should I go ahead and merge this?

PR #57, opened 2019-11-06
Add an exit/quit command for sdb

ctuffli:
At the OpenZFS hackathon, someone asked how to exit the debugger. This commit adds an explicit `exit` command in addition to the current `Ctrl-D`.

### DIFF
--- a/sdb/commands/exit.py
+++ b/sdb/commands/exit.py
@@ -16,7 +16,7 @@
 
 # pylint: disable=missing-docstring
 
-import argparse
+import sys
 from typing import Iterable
 
 import drgn
@@ -29,4 +29,4 @@ class Echo(sdb.Command):
     names = ["exit", "quit"]
 
     def call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
-        exit()
+        sys.exit()
--- a/sdb/commands/exit.py
+++ b/sdb/commands/exit.py
@@ -16,7 +16,6 @@
 
 # pylint: disable=missing-docstring
 
-import sys
 from typing import Iterable
 
 import drgn
@@ -29,4 +28,4 @@ class Echo(sdb.Command):
     names = ["exit", "quit"]
 
     def call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
-        sys.exit()
+        raise SystemExit
--- a/sdb/commands/exit.py
+++ b/sdb/commands/exit.py
@@ -1,0 +1,32 @@
+#
+# Copyright 2019 Chuck Tuffli
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pylint: disable=missing-docstring
+
+import argparse
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class Echo(sdb.Command):
+    # pylint: disable=too-few-public-methods
+
+    names = ["exit", "quit"]
+
+    def call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
+        exit()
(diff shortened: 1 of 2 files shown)


sdb/commands/exit.py
```diff
@@ -0,0 +1,32 @@
...
+import sdb
+
+
+class Echo(sdb.Command):
+    # pylint: disable=too-few-public-methods
+
+    names = ["exit", "quit"]
+
+    def call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
+        sys.exit()
```
prakashsurya: Should we use 'raise SystemExit` instead? I'm not fluent in python enough to explain way the exception may be better, but I've seen the exception recommended in other code bases, so I figured I'd at least ask.

sdb/commands/exit.py
```diff
@@ -0,0 +1,32 @@
...
+import drgn
+import sdb
+
+
+class Echo(sdb.Command):
+    # pylint: disable=too-few-public-methods
+
+    names = ["exit", "quit"]
+
+    def call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
```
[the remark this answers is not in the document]
sdimitro: I'm ok with this code staying as is.

Thinking out loud about this, at some point we'll want to enable `mypy` type checks as part of the test infrastructure. I'm sure before the commit that enables that kind of testing, we'll have a commit fixing all the type annotations to make the type checks happy. With that in mind, I'd encourage breaking out of the norm if we want to be more precise like applying the `-> None` change that Prakash suggests. This way we can really test ourselves later how effective is `mypy` at finding issues that we really care about, or whether it is just a tool that looks at superficial things.

In the above case it's not like we'd be lying anyway. Setting the return value as `None` seems more precise.

sdb/commands/exit.py
sdimitro: Doing some superficial research it seems like `sys.exit()` and `raise SystemExit` do the same thing.

In my opinion though I think it would be preferable to `raise SystemExit` and then catch it in `sdb/internal/repl.py` here:
```
def start_session(self) -> None:
        """
        Starts a REPL session.
        """
        while True:
            try:
                line = input(self.prompt).strip()
            except (EOFError, KeyboardInterrupt, SystemExit): # <----- Added here
                print(self.closing)
                break

            if not line:
                continue
            _ = self.eval_cmd(line)
```

My reasoning for this has two points:
1] All the exit code paths (currently Ctrl+D and the `exit` command) go through the same closing code of the session of our REPL. This is helpful because, if we were to do add more cleanup functionality, you'd only have to care about this code path, not multiple codepaths (e.g. one for Ctrl+D and one for `exit`).
2] In the wild scenario when we have another consumer outside of the REPL (e.g. an SDB server that we contact over the network or whatever) we decouple that functionality by raising an Exception to exit and each consumer can handle exit however they want. Doing `sys.exit` and instantly halt the program may not be what each consumer as context may vary.

That said, I do find it annoying that we'd have to use an exception for something that is not very *exceptional* such as exiting the session. That said, I believe that the above benefits outweigh the drawbacks.

prakashsurya: I think the ability to catch the exit exception, if the consumer wishes to, is the main benefit of using it. Those two points specific to SDB make sense to me.

ctuffli: I'm not a Python expert but checked with someone here, and they agreed that there would be a benefit to using `raise SystemExit` for the reason you both mentioned. I'll make that change and update a bit later.

sdimitro:
Hey @ctuffli , the code looks good to me!

Could you please squash all your commits together into a single commit before I merge this?
If you don't have time, I can probably do that for you later this week and push that commit with you as the author.

ahrens:
@sdimitro FYI, if we enable "Squash and merge" for this repo, then you can easily squash it yourself from the Github BUI while merging the change.  (we may need to ask an admin like @eschrock to do this)

prakashsurya:
@ahrens we can enable that, but how will that work with the merge commits?

prakashsurya:
I've enabled the option, but I'm hesitant to use it since I'm not very familiar with it.

ahrens:
@prakashsurya I haven't used it on PR's that contain merge commits before, but I'd hope that it rebases the whole thing as a single commit.  I'd say give it a shot and we'll see :)

sdb/commands/exit.py
```diff
@@ -0,0 +1,31 @@
...
+
+# pylint: disable=missing-docstring
+
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class Echo(sdb.Command):
```
ahrens: I realize this is too late, since this was just integrated, but I think the convention is for the class name to be related to the file name, i.e. `Exit` in this case.

ctuffli: Good catch. New PR on its way.

PR #62, opened 2019-11-06
Fix argparse crash when opening dump from the same system

### LINKED ISSUE #21: argparse crash when trying to open a dump from the same system that generated it
Matt (@ahrens) hit the following issue:
```
$ sudo sdb /boot/vmlinuz-4.15.0-65-generic dump.201910042034
  File "/usr/bin/sdb", line 11, in <module>
  File "/usr/lib/python3/dist-packages/sdb/internal/cli.py", line 206, in main
    prog = setup_target(args)
  File "/usr/lib/python3/dist-packages/sdb/internal/cli.py", line 162, in setup_target
    args.symbol_search.insert(0, args.object)
AttributeError: 'NoneType' object has no attribute 'insert'
```

The problem is that `args.symbol_search` was not initialized because the `-s` option was not specified (we don't expect to use `-s` if we are opening the crash dump from the same system).

The current workaround is to do something like this:
```
sudo sdb /boot/vmlinuz-4.15.0-65-generic dump.201910042034 -s /boot/vmlinuz-4.15.0-65-generic
```

Matt raised a good point where it would be nice for `sdb` to guess at where the kernel image and modules are when given a crash dump. We should definitely try to implement this while fixing this bug.

sdimitro:
= Bug Description

There is a bug when trying to open a crash dump from the same
system that generated it (e.g. the user skips specifying where
the kernel module directories are with the -s option) crashes
SDB at startup.

= Root-cause

Looking at the actual failure below:
```
$ sudo sdb /boot/vmlinuz-4.15.0-65-generic dump.201910042034
  File "/usr/bin/sdb", line 11, in <module>
  File "/usr/lib/python3/dist-packages/sdb/internal/cli.py", line 206, in main
    prog = setup_target(args)
  File "/usr/lib/python3/dist-packages/sdb/internal/cli.py", line 162, in setup_target
    args.symbol_search.insert(0, args.object)
AttributeError: 'NoneType' object has no attribute 'insert'
```

The problem is that we expect `args.symbol_search` to exist as
a variable and we add the path of vmlinux to it (first argument
from the CLI) so drgn can import all the symbols and DWARF info.
That variable doesn't exist if the `-s` option is not used, thus
we fail with the above error when that option is not present.

= Fix

Ensure that this variable is set anyways even if `-s` is not
present.

= Testing

Before the patch:
```
$ sudo sdb  vmlinux-4.15.0-58-generic 201908192231/dump.201908192231
  File "/usr/local/bin/sdb", line 11, in <module>
  File "/usr/local/lib/python3.6/dist-packages/sdb-0.1.0-py3.6.egg/sdb/internal/cli.py", line 206, in main
    prog = setup_target(args)
  File "/usr/local/lib/python3.6/dist-packages/sdb-0.1.0-py3.6.egg/sdb/internal/cli.py", line 162, in setup_target
    args.symbol_search.insert(0, args.object)
AttributeError: 'NoneType' object has no attribute 'insert'
```

After the patch:
```
$ sudo sdb  vmlinux-4.15.0-58-generic 201908192231/dump.201908192231
>
```

Closes #21

### DIFF
--- a/sdb/internal/cli.py
+++ b/sdb/internal/cli.py
@@ -66,7 +66,7 @@ def parse_arguments() -> argparse.Namespace:
         "-s",
         "--symbol-search",
         metavar="PATH",
-        type=str,
+        default=[],
         action="append",
         help="load debug info and symbols from the given directory or file;" +
         " this may option may be given more than once",
@@ -120,6 +120,7 @@ def parse_arguments() -> argparse.Namespace:
     #
     if args.object and not args.core:
         parser.error("raw object file target is not supported yet")
+
     return args
 
 
@@ -159,7 +160,7 @@ def setup_target(args: argparse.Namespace) -> drgn.Program:
         # or userland binary using the non-default debug info
         # load API.
         #
-        args.symbol_search.insert(0, args.object)
+        args.symbol_search = [args.object] + args.symbol_search
     elif args.pid:
         prog.set_pid(args.pid)
     else:


sdb/internal/cli.py
```diff
@@ -120,6 +120,15 @@ def parse_arguments() -> argparse.Namespace:
...
     if args.object and not args.core:
         parser.error("raw object file target is not supported yet")
+
+    #
+    # For the case where the -s option is not specified (e.g.
+    # sdb <binary> <dump>) we instantiate symbol_search ourselves
+    # so it has a type even though it is an empty list.
+    #
+    if not args.symbol_search:
+        args.symbol_search = []
```
prakashsurya: When adding the "symbol-search" option above, it looks like we define it to be a "str", so it strikes me as odd that we'd initialize it to an empty list here, rather than an empty string..

Can you comment on why the types differ? Can we consistently use the same type for both cases?

Also, could we more simply set a "default" value when one is not specified on the command like, using argparse directly? i.e. use `default=...` when we create/add that argument via `dis_group.add_argument`.

sdimitro: You are right, I'll give it another look soon.

sdimitro:
@prakashsurya can you take another look when you have the time.
Decided to remove the type annotation and set a default which implies the type anyway.
Changed the use of `insert` to something more generic that could also handle the case when that variable is set to `None`

PR #70, opened 2019-11-14
Introduce new "help" command

prakashsurya:
This change adds a new "help" command that can be used to pring out
useful "help messages" for any particular SDB command. The contents of
the help message is derived automatically from the command class's
docstring and argument parser.

Co-authored-by: Prakash Surya <prakash.surya@delphix.com>

### DIFF
--- a/sdb/command.py
+++ b/sdb/command.py
@@ -66,7 +66,7 @@ class Command:
             #
             if i == 0:
                 line = line.replace('usage: ', '')
-            print("    {}".format(line.replace('usage: ', '')))
+            print("    {}".format(line))
 
         if len(cls.names) > 1:
             print("ALIASES")
@@ -81,7 +81,7 @@ class Command:
         #
         if cls.__doc__:
             #
-            # The first line of the docstring is the summary, whichis
+            # The first line of the docstring is the summary, which is
             # already be included in the parser description. The second
             # line should be empty. Thus, we skip these two lines.
             #
--- a/sdb/command.py
+++ b/sdb/command.py
@@ -34,21 +34,67 @@ class Command:
 
     # pylint: disable=too-few-public-methods
 
+    @classmethod
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        #
+        # If the class doesn't have a docstring, "inspect.getdoc" will
+        # pull from the parent class's docstring, if the parent class
+        # does have a docstring. This is not the behavior we want, so we
+        # first check "cls.__doc__" before using "inspect.getdoc".
+        #
+        if cls.__doc__:
+            summary = inspect.getdoc(cls).splitlines()[0].strip()
+        else:
+            summary = None
+        return argparse.ArgumentParser(prog=name, description=summary)
+
+    @classmethod
+    def help(cls, name: str):
+        """
+        This method will print a "help message" for the particular
+        command class that it's called on. The docstring and parser for
+        the class is used to populate the contents of the message.
+        """
+        parser = cls._init_parser(name)
+
+        print("SUMMARY")
+        for i, line in enumerate(parser.format_help().split('\n')):
+            #
+            # When printing the help message, the first line will have a
+            # "usage: " prefix string which looks awkward, so we strip
+            # that prefix prior to printing the first line.
+            #
+            if i == 0:
+                line = line.replace('usage: ', '')
+            print("    {}".format(line.replace('usage: ', '')))
+
+        if len(cls.names) > 1:
+            print("ALIASES")
+            print("    {}".format(", ".join(cls.names)))
+            print()
+
+        #
+        # If the class doesn't have a docstring, "inspect.getdoc" will
+        # pull from the parent class's docstring, if the parent class
+        # does have a docstring. This is not the behavior we want, so we
+        # first check "cls.__doc__" before using "inspect.getdoc".
+        #
+        if cls.__doc__:
+            #
+            # The first line of the docstring is the summary, whichis
+            # already be included in the parser description. The second
+            # line should be empty. Thus, we skip these two lines.
+            #
+            for line in inspect.getdoc(cls).splitlines()[2:]:
+                print("{}".format(line))
+            print()
+
     #
     # names:
     #    The potential names that can be used to invoke
     #    the command.
     #
     names: List[str] = []
-
-    #
-    # name:
-    #    The name used when the command was invoked. This
-    #    is generally used as a parameter passed to
-    #    exceptions raised by SDB to make error messages
-    #    more precise.
-    #
-    name: str = ""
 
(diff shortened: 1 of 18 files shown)


prakashsurya:
cc: @sdimitro @ctuffli @ahrens 

This is a continuation/extention of: https://github.com/delphix/sdb/pull/68 .. I've taken what @ctuffli originally proposed, and tried to get it into a state that can be merged. This PR has the following differences from what was originally proposed:

- I've removed the `-v` flag to the `help` command; now it always emits verbose output
- I've removed the ability to run `help` without specifying a command; I think we should revisit this in a follow up PR/bug, to determine how to address the "what commands exist?" question raised in the original PR
- I've integrated with each command's parser, so we get all of the argparse information in the help message (e.g. description, options, etc)
- I refactored how the parser object gets created, so that we can create it in a `@classmethod` without instantiating a new object
- I removed the changes that add new docstrings to the commands; I'd like to do this in a seperate commit

As is, the format of the help message is the following:
```
> help spa
SUMMARY
    spa [-h] [-v] [-m] [-H] [-w] [poolnames [poolnames ...]]

    positional arguments:
      poolnames

    optional arguments:
      -h, --help       show this help message and exit
      -v, --vdevs      vdevs flag
      -m, --metaslab   metaslab flag
      -H, --histogram  histogram flag
      -w, --weight     weight flag

```
```
> help pp
SUMMARY
    pp [-h]

    optional arguments:
      -h, --help  show this help message and exit

ALIASES
    pretty_print, pp

```

Once we add docstrings to certain commands, I expect this to expand into something like this:
```
> help head                                                                                                          
SUMMARY                                                                                                       
    head [-h] [count]                                                                                            
                                                                                                                 
    Return the first N objects passed in the pipeline                                                                 
                                                                                                                       
    positional arguments:                                                                                   
      count                                                                                                              
                                                                                                                         
    optional arguments:                                                                                               
      -h, --help  show this help message and exit                                                                    
                                                                                                                       
DESCRIPTION                                                                                                 
                                                                                                                    
    Command takes a 'count' parameter which defaults to 10                                                       
                                                                                                                         
EXAMPLES                                                                                                             
                                                                                                                      
    > spa | head 1 | member spa_name                                                                                  
    (char [256])"domain0"  

```

Please let me know what you all think about what I have here. If we're in agreement with this approach, I think we can land this, and the circle back to the following items:

- Add new docstrings to each command
- Determine how to answer the "what commands are available?" question

sdb/command.py
```diff
@@ -34,22 +34,68 @@ class Command:
...
+        #
+        # If the class doesn't have a docstring, "inspect.getdoc" will
+        # pull from the parent class's docstring, if the parent class
+        # does have a docstring. This is not the behavior we want, so we
+        # first check "cls.__doc__" before using "inspect.getdoc".
+        #
+        if cls.__doc__:
+            summary = inspect.getdoc(cls).splitlines()[0].strip()
+        else:
+            summary = None
```
sdimitro: Nice!

sdb/command.py
```diff
@@ -34,22 +34,68 @@ class Command:
...
 
-    #
-    # name:
-    #    The name used when the command was invoked. This
-    #    is generally used as a parameter passed to
-    #    exceptions raised by SDB to make error messages
-    #    more precise.
-    #
-    name: str = ""
-
```
sdimitro: Out of curiosity, why is this removed? Asking because I liked the comment making the differentiation between `name` and `names`.

sdb/commands/address.py
```diff
@@ -42,9 +42,11 @@ class Address(sdb.Command):
 
     names = ["address", "addr"]
 
-    def _init_argparse(self, parser: argparse.ArgumentParser) -> None:
-        super()._init_argparse(parser)
+    @classmethod
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        parser = super(Address, cls)._init_parser(name)
```
sdimitro: It's a bit annoying that we have to do it like that, is there a reason why the following doesn't work:
```
parser = super()._init_parser(name)
```
Also if you actually gave the above a try is there any advantage on using the more verbose version that you have?

sdb/command.py
```diff
@@ -34,22 +34,68 @@ class Command:
...
+        print("SUMMARY")
+        for i, line in enumerate(parser.format_help().split('\n')):
+            #
+            # When printing the help message, the first line will have a
+            # "usage: " prefix string which looks awkward, so we strip
+            # that prefix prior to printing the first line.
+            #
+            if i == 0:
+                line = line.replace('usage: ', '')
+            print("    {}".format(line.replace('usage: ', '')))
```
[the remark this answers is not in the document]
prakashsurya: Sorry, good catch. I meant to remove the second replace (which I used first, when iterating).

While perhaps we could remove it on every line, I wanted to be more targeted and only remove it on the first line.

I'll fix this.

sdb/commands/address.py
prakashsurya: I can try that; I can't recall if I had tried that or not. This is a class method, though, so I think we need to do it this way. But I'll double check.. I also would much rather just use `super()` if possible.

sdb/command.py
prakashsurya: So, based on my understanding of python, this will create a class level variable named `name`, which is different than the object level variable named `name`. I didn't see where we used the class level `name` variable, vs. how we do use the object level `self.name` variable.

Assuming I'm correct and this is dead code, I figured it should be removed to avoid confusion.

Let me know if I'm understanding this incorrectly.

sdb/commands/address.py
prakashsurya: It looks like `super()` does work, so I don't know why I didn't try that to begin with. I guess my lack of experience with Python is showing :)

sdb/command.py
sdimitro: I see. I honestly had no idea that this was a thing. I'm resolving this.

prakashsurya:
Using the github UI to squash and merge unfortunately lost the `Author` attribution to @ctuffli:
```
commit ca9d5c2d668f9bff66593d75ee351a24ce28bfc0
Author: Prakash Surya <prakash.surya@delphix.com>
Date:   Fri Nov 15 12:50:46 2019 -0800

    Introduce new "help" command (#70)
    
    This change adds a new "help" command that can be used to print out
    useful "help messages" for any particular SDB command. The contents of
    the help message is derived automatically from the command class's
    docstring and argument parser.
    
    Co-authored-by: Prakash Surya <prakash.surya@delphix.com>
```
@sdimitro should we attempt to rectify this, or just leave it as is? I'm not sure we can do anything, without a force push, and changing the commit history of the repository...

ctuffli:
Don't sweat the attribution on this

Issue #28, opened 2019-10-07
Want `btree` walker for ZFS


sdimitro:
To be used for range trees mostly

sdimitro:
cc @shartse

PR #71, opened 2019-11-15
Add btree walker

shartse:
Add btree walker: https://github.com/delphix/sdb/issues/28. This is based on ZFS's btree implementation: https://github.com/zfsonlinux/zfs/blob/master/include/sys/btree.h

I've tested it on `ms_allocatable` trees. I was able to verify that it finds the correct number of elements:
```
*(zfs_btree_t *)0xffff9268a76e3800 = {
	.bt_root = (zfs_btree_hdr_t *)0xffff9268a89fa000,
	.bt_height = (int64_t)1,
	.bt_elem_size = (size_t)8,
	.bt_num_elems = (uint64_t)2436,
	.bt_num_nodes = (uint64_t)8,
	.bt_bulk = (zfs_btree_leaf_t *)0x0,
	.bt_compar = (int (*)(const void *, const void *))0xffffffffc0465b50,
}
> echo 0xffff9268a76e3800 | cast zfs_btree_t * | walk | cast range_seg32_t * |count
(unsigned long long)2436
```
And, based on visual inspection, that it produces elements in sorted order:
```
 echo 0xffff9268a76e3800 | cast zfs_btree_t * | walk | cast range_seg32_t * 
*(range_seg32_t *)0xffff9268a711d010 = {
	.rs_start = (uint32_t)129111,
	.rs_end = (uint32_t)129119,
}
*(range_seg32_t *)0xffff9268a711d018 = {
	.rs_start = (uint32_t)129143,
	.rs_end = (uint32_t)129157,
}
...
*(range_seg32_t *)0xffff92672dfd5d58 = {
	.rs_start = (uint32_t)1048346,
	.rs_end = (uint32_t)1048576,
}
```
I'm open to other suggestions on how to test this, especially on trees with greater depth.

### DIFF
--- a/sdb/commands/zfs/btree.py
+++ b/sdb/commands/zfs/btree.py
@@ -1,0 +1,85 @@
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pylint: disable=missing-docstring
+
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class Btree(sdb.Walker):
+    """
+    walk zfs_btree
+
+    DESCRIPTION
+
+        btree_t's are used in ZFS, especially as metaslab range trees. Use walk or
+        zfs_btree for an in-order traversal of the contents of the tree.
+
+    EXAMPLES
+
+        > <metaslab_t *> | member ms_allocatable | walk | cast range_seg32_t *
+
+        *(range_seg32_t *)0xffff9990b5141d68 = {
+        .rs_start = (uint32_t)862981,
+        .rs_end = (uint32_t)868158,
+        }
+        *(range_seg32_t *)0xffff9990b5141d70 = {
+        .rs_start = (uint32_t)868163,
+        .rs_end = (uint32_t)939419,
+        }
+        *(range_seg32_t *)0xffff9990b5141d78 = {
+        .rs_start = (uint32_t)939424,
+        .rs_end = (uint32_t)1048576,
+        }
+    """
+
+    names = ["zfs_btree"]
+    input_type = "zfs_btree_t *"
+
+    def __init__(self, prog: drgn.Program, args: str = "",
+                 name: str = "_") -> None:
+        super().__init__(prog, args, name)
+        self.elem_size = None
+
+    def _val(self, start: int, idx: int) -> drgn.Object:
+        location = start + (self.elem_size * idx)
+        return drgn.Object(self.prog, type="void *", value=location)
+
+    def _helper(self, node: drgn.Object) -> Iterable[drgn.Object]:
+        if not node:
+            return
+
+        count = node.bth_count
+        if node.bth_core:
+            # alterate recursive descent on the children and generating core objects
+            core = drgn.cast('struct zfs_btree_core *', node)
+            for i in range(count):
+                yield from self._helper(core.btc_children[i])
+                yield self._val(core.btc_elems, i)
+            # descend the final, far-right child node
+            yield from self._helper(core.btc_children[count])
+        else:
+            # generate each object in the leaf elements
+            leaf = drgn.cast('struct zfs_btree_leaf *', node)
+            for i in range(count):
+                yield self._val(leaf.btl_elems, i)
+
+    def walk(self, obj: drgn.Object) -> Iterable[drgn.Object]:
+        self.elem_size = obj.bt_elem_size
+        yield from self._helper(obj.bt_root)


sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,56 @@
...
+    """ walk  btree """
+
+    names = ["zfs_btree"]
+    input_type = "zfs_btree_t *"
+
+    def _val(self, start, size, idx):
+        return start + (size * idx)
+
+    def _helper(self, node: drgn.Object, elem_size : int) -> Iterable[drgn.Object]:
+        if node == drgn.NULL(self.prog, node.type_):
```
ahrens: I think that NULL pointers compare as false, so you can do `if not node:`

sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class Btree(sdb.Walker):
+    """ walk  btree """
+
+    names = ["zfs_btree"]
+    input_type = "zfs_btree_t *"
+
+    def _val(self, start, size, idx):
+        return start + (size * idx)
```
ahrens: I think it would be even more useful to have this function return the drgn.Object at this address, since that's all its return value is used for currently.

sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def _helper(self, node: drgn.Object, elem_size : int) -> Iterable[drgn.Object]:
```
ahrens: since `elem_size` is constant for this invocation, and it's used in all the helper functions, would it make sense to store it in an instance variable, rather than passing it around everywhere?

sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,56 @@
...
+import sdb
+
+
+class Btree(sdb.Walker):
+    """ walk  btree """
+
+    names = ["zfs_btree"]
+    input_type = "zfs_btree_t *"
+
+    def _val(self, start, size, idx):
```
sdimitro: nit: regardless of what you decide to do with Matt's comment below could you add types to this method?

sdimitro:
Also looking at the check failures it seems like `lint` gave the error below:
```
Run pylint -d duplicate-code sdb
************* Module sdb.commands.zfs.btree
sdb/commands/zfs/btree.py:34:51: C0326: No space allowed before :
    def _helper(self, node: drgn.Object, elem_size : int) -> Iterable[drgn.Object]:
                                                   ^ (bad-whitespace)
sdb/commands/zfs/btree.py:31:4: R0201: Method could be a function (no-self-use)

-----------------------------------
Your code has been rated at 9.99/10
```
and `yapf` a styling error at the same part:
```
Run yapf --diff --style google --recursive sdb
--- sdb/commands/zfs/btree.py	(original)
+++ sdb/commands/zfs/btree.py	(reformatted)
@@ -31,7 +31,8 @@
     def _val(self, start, size, idx):
         return start + (size * idx)
 
-    def _helper(self, node: drgn.Object, elem_size : int) -> Iterable[drgn.Object]:
+    def _helper(self, node: drgn.Object,
+                elem_size: int) -> Iterable[drgn.Object]:
         if node == drgn.NULL(self.prog, node.type_):
             return
 
##[error]Process completed with exit code 1.
```

==
The `lint` error basically is asking to convert that method to a function like this:
```
@staticmethod
def _helper(node: drgn.Object, elem_size : int) -> Iterable[drgn.Object]:
... ^notice that self was removed from the arguments
```
you can either do the above OR if you end up taking Matt's advice about making `elem_size` global you will end up using `self.elem_size` in this method making the lint error go away.

As for the styling error you can install `yapf` and make it format the code correctly for you:
```
python3 -m pip install yapf
yapf -r -i --style google sdb
```

sdb/commands/zfs/btree.py
shartse: Yeah - I think that makes sense. I added it as an instance variable, let me know if you think I can improve the style - my python's pretty rusty.

sdb/commands/zfs/btree.py
shartse: Yup will do. Another thing I wanted to ask was whether I need to transform more of the Drgn objects I'm using into `int`s (like `elem_size`)? It seems to work as is, but I'm not sure about the correctness or best practices.

sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,60 @@
...
+# pylint: disable=missing-docstring
+
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class Btree(sdb.Walker):
+    """ walk  zfs_btree """
```
prakashsurya: We just integrated the new `help` command. It'd be great if we could use this PR as an opportunity to add documentation that could be used as an example for other commands. For example, we we could use a format like this:
```
"""
walk zfs_btree

DESCRIPTION

    A (potentially multiline) description about what the command does.

EXAMPLES

    > some_command | zfs_btree
    (example output)
```

I believe this will work as-is, and at least include a short description of the command (it'll use this one line summary); but if we could add more information that'd be great, and give us an example to use later for the rest of our commands.

sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,60 @@
...
+
+    names = ["zfs_btree"]
+    input_type = "zfs_btree_t *"
+
+    def __init__(self, prog: drgn.Program, **kwargs):
```
prakashsurya: Can we add the type annotation for the return value.. i.e. `-> None:`

Can you also add any information about the benefits/drawbacks of using `**kwargs` here, vs. how the other commands define the constructor? Should we (later) adopt using `**kwargs` for all other commands?

sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,60 @@
...
+        self.elem_size = None
+        super().__init__(prog, **kwargs)
+
+    def _val(self, start: int, idx: int):
```
prakashsurya: Can we add the type annotation for the return value.. i.e. `-> drgn.Object:`

sdb/commands/zfs/btree.py
shartse: Sure - I'll give it try!

sdb/commands/zfs/btree.py
prakashsurya: > Can you also add any information about the benefits/drawbacks ...
To be clear (after re-reading this, I think it might not be), I just mean as a comment/reply in the PR.. not something to add in the code.. I'm just curious if we should migrate to using `**kwargs` across the board.

ahrens: In this case, I don't think the constructor (`__init__`) is needed at all.

shartse: @prakashsurya - I don't have a super strong intuition one way or the other - I was basing this on the implementation of `__init_subclass__` in `command.py`. I think having it match the other commands with `args: str = ""` makes sense to me. I think that's more explicit about what exactly the function has to work with. 

@ahrens - I added it to have a place to specify instance variables, but I guess since this is python, we don't actually need to do that.. Do you know if there's a stylistic imperative one way or the other? I guess I'm leaning towards getting rid of it, since we're not actually using it to instantiate `elem_size`.

ahrens: `__init__subclass__` is black magic.  You could model your `__init__` on other `__init__` methods, preferably of sibling classes (i.e. other subclasses of Walker).

AFAIK there isn't a convention to "declare" instance variables.  IMO it would be nice, though!

sdb/commands/zfs/btree.py
shartse: Alright - I wrote a simple help message. Let me know if you have any suggestions

sdb/commands/zfs/btree.py
shartse: Yeah that all makes sense to me. Interestingly, I just got this `pylint` warning:
```
sdb/commands/zfs/btree.py:79:8: W0201: Attribute 'elem_size' defined outside __init__ (attribute-defined-outside-init)
```

sdb/commands/zfs/btree.py
```diff
@@ -0,0 +1,85 @@
...
+	.rs_end = (uint32_t)868158,
+	}
+	*(range_seg32_t *)0xffff9990b5141d70 = {
+	.rs_start = (uint32_t)868163,
+	.rs_end = (uint32_t)939419,
+	}
+	*(range_seg32_t *)0xffff9990b5141d78 = {
+	.rs_start = (uint32_t)939424,
+	.rs_end = (uint32_t)1048576,
+	}
```
sdimitro: Out of curiosity do the tabs look aligned when the help message is printed?
From the code it seems like the first line in the description section is not aligned with the second one, similarly for the lines in the example section.

shartse: It looks fine in the output, but yeah, there's a tabs vs spaces issue. I'll update the PR.

PR #91, opened 2019-11-20
Remove TravisCI integration

prakashsurya:
Use of GitHub Actions is working well, so there's no need to continue to
use TravisCI. This change removes the TravisCI configuration.


prakashsurya:
> I'm just looking at the current PR where the TravisCI job seems to be running indefinitely

The issue is actually that it's not running, but we have the repository configured to require that test to run and pass. I'll update the repository settings to require the new Actions tests, instead of the Travis one.

prakashsurya:
@sdimitro I updated the repository settings, so it looks good now. I'll wait to merge this until you can give this another look (with the updated settings).

sdimitro:
@prakashsurya thanks for updating those. Everything looks good to me, feel free to merge whenever you want.

PR #92, opened 2019-11-20
Run Actions on a regular basis

prakashsurya:
Currently our tests are not entirely reproducible, since we don't pin
our dependencies to specific versions. Thus, frequently we'll open a
pull request, only to realize that a depdency has changed, and a new
test regression was silently introduced.

To help us notice these regressions sooner, this change is augmenting
our tests to run on a daily basis. This way, we'll continually test with
new versions of our dependencies, which should help us detect when
regressions are introduced, and allow us to fix the issue before the
test failure cascades into an unrelated pull request.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,9 @@
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron:  '0 0 * * *'
+
 jobs:
   #
   # Verify the build and installation of SDB.


prakashsurya:
@sdimitro I'm not entirely sure if I have the syntax correct for this; I tried to search around for an example, but didn't find anything. I'm also not sure how to test this, without landing it to master, and seeing what happens, unfortunately. I've never used the `schedule` feature before, but this is what I came up with after reading some of the docs ([here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onschedule)).

jgallag88:
@prakashsurya I'm not positive, but I think you might be able to test it by pushing it to the master branch of your fork.

prakashsurya:
@jgallag88 I've done that, using `* * * * *` as the cron string, and haven't seen any scheduled test runs yet.

prakashsurya:
Sorry, I see some `schedule` runs on my repository now, along with `push` runs, so I think this should work.

prakashsurya:
> [1] Who gets notified and how?

I'm not exactly sure who will get notified for this repository. For my personal fork, I get emails. IIRC, I also get emails when a PR of mine fails the tests. I haven't yet seen a test failure for a commit that we pushed to this repository, so I'm not sure how that'll work.

I was hoping to just land this, and see how it goes; worst case we can revert it. Since I don't have any experience with this yet, I'm learning as we go.

> [2] Does having a broken nightly introduce any new disruptions in our PR workflow?

I do not expect this to change PRs at all. I believe that if nightly catches a failure, we'll be able to open a PR to fix the issue, just like we currently do.

> [3] Would this infrastructure keep sending us emails for nightlies that fail every night even if no change has been applied to the repo? or will it just fail once and wait for something to happen?

I don't know for sure, but I'm expecting it to send a notification for each test failure; i.e. if it's broken for multiple days, I would expect to get multiple notifications.

PR #110, opened 2019-11-27
Reorganization and Cleanup of Classes

sdimitro:
The motivation of this commit is to prepare the repo for
automated static-type tests with mypy. The main thing
needed here was the fixing of cyclic imports in our
infrastructure that were confusing mypy.

= Fix existing cyclic imports issues

In our infrastrucure we declare the following hierarchy of
files:
[1] On the top level we have the modules: error.py and target.py
    (Keep in mind that for better or worst in Python each file
     is its own module). The target module is a very thin wrapper
    around the drgn.Program used by an SDB session.
[2] The command module uses the functionality provided by the
    above two modules. The command module (command.py) now
    includes the Locator, PrettyPrinter, and Walker classes
    as well as some very basic commands like Cast, Walk, and
    Address.
[3] The pipeline module uses the functionality provided by all
    of the above, mixing and matching commands to provide the
    pipeing functionality.
[4] The SDB module (e.g. sdb/__init__.py) imports all of the
    above and exposes them as its own so user-created modules
    can do things like `sdb.execute_pipeline(..)` after
    `import sdb` instead of having to do explicit imports of
    its internal infrastructure (e.g. `import sdb.pipeline`
    and `sdb.pipeline.execute_pipeline`).

= More help messages

Help messages for Cast, Address, and Walk.

= Pylint error cleanup

Things like too-few-public-methods are no longer a thing.


sdb/command.py
```diff
@@ -17,10 +17,38 @@
...
+    command will be available from the SDB REPL.
+    """
+    # pylint: disable=global-statement
+    global all_commands
+    all_commands[name] = class_
+
+
+def get_registered_commands() -> Dict[str, Type["Command"]]:
+    """
+    Return a dictionary of command names to command objects.
```
prakashsurya: should we be more specific and state "class objects" rather than "objects"? or perhaps just "class"? otherwise, I might interpret that to mean an instance object of the class, rather than the class itself.

sdb/command.py
```diff
@@ -113,59 +141,390 @@ def __init_subclass__(cls, **kwargs):
...
+            except TypeError as err:
+                raise CommandError(self.name, str(err))
+
+
+class Address(Command):
+    """
+    Return address of the given object
+
+    DESCRIPTION
+        The command accepts input both from the pipe and the SDB CLI.
```
prakashsurya: "input both from" -> "input from both"

sdb/command.py
```diff
@@ -113,59 +141,390 @@ def __init_subclass__(cls, **kwargs):
...
+
+        For objects passed from the pipe, their address is returned
+        only if they are legitimate objects created from the target
+        program being examined. If this objects were created on the
```
prakashsurya: "If this objects" -> "If the objects"

sdb/command.py
```diff
@@ -113,59 +141,390 @@ def __init_subclass__(cls, **kwargs):
...
+        fly with something like echo and thus don't have an actual
+        address in the address space of the program being examined,
+        this command just passes on their values as pointers.
+
+        The user can specify one or more inputs as argument to the
```
prakashsurya: "argument" -> "arguments"

sdb/command.py
```diff
@@ -113,59 +141,390 @@ def __init_subclass__(cls, **kwargs):
...
+            register_command(name, cls)
 
     def call(self,
              objs: Iterable[drgn.Object]) -> Optional[Iterable[drgn.Object]]:
         # pylint: disable=missing-docstring
         raise NotImplementedError
 
-    def _call_and_yield(self,
-                        objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
+    def call_and_yield(self,
```
prakashsurya: While you're making lots of changes, would it be reasonable to rename this to `call`, and rename the existing `call` to `_call` (or `_call_impl`, or something)?

Matt and I briefly discussed doing something like this.. so external consumers would use `call` and that would always return an iterable.. rather than having two external method, `call` and `call_and_yield`?

sdb/command.py
```diff
@@ -113,59 +141,390 @@ def __init_subclass__(cls, **kwargs):
...
+
+class Walk(Command):
+    """
+    Dispatch the appropriate walker based on the type of input
+
+    DESCRIPTION
+        This command can be used to walk data structures when
+        a specific walker for them already exists. There are
+        two scenarios when this command is preferable to using
+        a speicific walker:
```
ahrens: typo: should be `specific`

sdb/commands/linux/dmesg.py
```diff
@@ -23,18 +23,17 @@
...
 
     def no_input(self) -> Iterable[drgn.Object]:
-        log_idx = sdb.prog["log_first_idx"]
-        log_seq = sdb.prog["clear_seq"]
-        log_end = sdb.prog["log_next_seq"]
-        log_buf = sdb.prog["log_buf"]
+        log_idx = sdb.get_object("log_first_idx")
+        log_seq = sdb.get_object("clear_seq")
+        log_end = sdb.get_object("log_next_seq")
+        log_buf = sdb.get_object("log_buf")
```
ahrens: how about renaming `get_object` to `get_symbol`?

sdb/target.py
```diff
@@ -0,0 +1,87 @@
...
+context of a running SDB instance. Its current form can be
+thought of as a singleton class where the infrastructure will
+initialize the `prog` global once and the consumers of the
+infrastructure will be using the helper methods defined in
+it to query information from the target. That said, there are
+still edge cases when consumers may still want the actual
+drgn.Program object (e.g. they are using the stack_trace() or
+list_for_each() helpers of the drgn API) and since global
+variables cannot explicitly imported from other modules without
+the star (*) notation, we provide the get_prog() wrapper.
```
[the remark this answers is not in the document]
sdimitro: `prog` is defined in `target.py` but we want to expose it to consumers through the `sdb` package (e.g. in `sdb/__init__.py` which is imported every time a new file does `import sdb`).

In other words we would like new commands to just use `import sdb` instead of
```
import sdb
import sdb.target
```
In order to better avoid import cycles in the future and provide a simpler API.

Now if we do the following in `sdb/__init__.py`
```
from sdb.target import prog
```
That `prog` is not re-exported from that module, so consumers cannot just access `prog` by doing `import sdb` or `from sdb import prog`. There are a couple of ways of re-exporting a variable like this through `__init__.py`:
```
# [1] The most common that is mentioned in the comment but also the ugliest/hackiest one because it imports everything
from sdb.target import *
# [2] The "redundant" one - I can switch the code to that if people feel it is better
from sdb.target import prog as prog
# [3] The most explicit one (reads as "this module's namespace consists of prog" - I can switch the code to that too if people feel it is better
from sdb.target import prog
__all__ = ['prog']
```
After thinking hard about the above though, I realized that maybe if I just expose an API of `sdb.target` to `sdb` that can get re-exposed it would lead to what I think is a more modular solution.

Does this answer your question? Also, can you think of a better way to write the comment that you've highlighted above?

sdb/command.py
sdimitro: Makes sense to me

sdb/command.py
sdimitro: Makes sense will do that

sdb/commands/linux/dmesg.py
sdimitro: Unfortunately, if you look at `target.py` you will see the following:
```
def get_object(obj_name: str) -> drgn.Object:
    global prog
    return prog[obj_name]


def get_symbol(sym: Union[str, int]) -> drgn.Symbol:
    global prog
    return prog.symbol(sym)
```
One returns a `drgn.Object` the other one a `drgn.Symbol` and their argument is slightly different (e.g. you can get a symbol by its address but not an object) - I decided to have them modeled after what they return from `drgn` as so far `target.py` is mostly wrappers around it. If you think we could do a better job and combine the above too or just rename them to something more precise or intuitive I'm open to ideas!

sdb/target.py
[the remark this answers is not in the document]
prakashsurya: Yea, so I guess the take away is, we don't want folks to use `sdb.target` directly?

e.g. while they could do this (which is what I was expecting they would/could do, to get access to `prog`):
```
from sdb.target import prog
```
in the specific command.. we'd rather them not, and just rely on:
```
import sdb
```
as the "public interface".. does that sound right? where, they'd then get access to the `get_prog` function..

I think adding some commentary about the fact that we don't want consumers to use `sdb.target` directly, would help explain the situation.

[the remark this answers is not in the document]
sdimitro: yeah that's correct. ok I will change the comment accordingly

sdimitro:
@ahrens @prakashsurya - can you take another look at the above? I think I resolved all the issues but the shipits were dismissed.

PR #111, opened 2019-11-28
Add help text for the basic commands

ctuffli:
Signed-off-by: Chuck Tuffli <ctuffli@gmail.com>


ctuffli:
I made a pass at what the commands do but need help in a few cases (e.g. pretty_print, pyfilter). I could also use some feedback on the examples for each command.

ctuffli:
I'm not that familiar with yapf may be misinterpreting the output, but it appears to be complaining about code I didn't touch. Do I need to reformat the code it flagged?

prakashsurya:
@ctuffli I've opened https://github.com/delphix/sdb/pull/112 to address the "yapf" errors; we can fix those in that PR, and leave this PR isolated to your proposed changes.

sdimitro:
Hey @ctuffli sorry for the late notice in this.

Prakash is right to point out that there is some duplicate work here with my aforementioned PR - I think you can skip doing the ones that I've already done in there (I think those are `Address`, `Walk`, and `Cast`) - except if you think they are not that good at which point feel free to open an issue at that review.

Thank you for working on this btw - the issues that you've raised and you're PRs have been important on improving the usability of `sdb`.

sdb/commands/count.py
```diff
@@ -23,6 +23,22 @@
...
+
+            sdb> addr 0 | count
+            (unsigned long long)1
+            sdb> addr 0 | addr 1 | count
+            (unsigned long long)2
+
+        Print the number of dbufs
+
+            sdb> dbuf | count
+            (unsigned long long)19
```
sdimitro: That's a good example but `dbufs` is a very generic term for someone who is not familiar with ZFS.
Maybe change the description to say `Print the number of ZFS dbufs`.

Other examples that are more generic for Linux folks are:
```
    Print the number of root slab caches in the system

        sdb> slabs | count
        (unsigned long long)136

    Count the number of level 3 log statements in the kernel log buffer.

        sdb> dmesg | filter obj.level == 3 | count
        (unsigned long long)24
```

sdb/commands/echo.py
```diff
@@ -24,6 +24,9 @@
 
 
 class Echo(sdb.Command):
+    """
+    ???
+    """
```
sdimitro: `echo` is really the same as `address` but without the symbol-to-address resolution and the ability to get addresses from `drgn.Object`s passed through a pipe. It just forwards whatever it is passed from the pipe as is and adds to the pipe void* objects created by whatever address is passed as an argument.

In other words, `echo` in the beginning of the pipeline is like a crippled `address` command. But in the middle of a pipeline, it acts differently as it can add objects to an existing pipeline without bothering any pre-existing pipeline objects.

sdb/commands/filter.py
```diff
@@ -24,6 +24,22 @@
...
+
+            sdb> addr 0 1 2 3 4 5 6 | filter obj >= 4
+            (void *)0x4
+            (void *)0x5
+            (void *)0x6
+
+        Print only pool "jax"
+
+            sdb> spa | filter obj.spa_name == "jax" | member spa_name
+            (char [256])"jax"
```
sdimitro: As noted previously, let's point out that this is ZFS specific by describing the example with `Find the SPA object of the ZFS pool named "jax" and print its 'spa_name'` or something along these lines.

In addition you can also re-use the `dmesg | filter | count` example from my other comment.

sdb/commands/stacks.py
```diff
@@ -52,6 +52,9 @@
 # future iterations.
 #
 class Stacks(sdb.Command):
+    """
+    Print the stack traces for the current threads
```
sdimitro: I think it would be great if we had some examples of using `stacks -m module` and `stacks -c function` here!

sdb/commands/echo.py
ctuffli: OK, how does this sound:
```
Creates an object from the given address or addresses
```
I went back through the presentation and tests but didn't really see a good example

sdb/commands/stacks.py
ctuffli: OK, I'll work on that

sdb/commands/echo.py
sdimitro: Sounds good to me - you may want add that the type of this object is a `void *`

sdb/commands/head.py
```diff
@@ -24,6 +24,15 @@
 
 
 class Head(sdb.Command):
+    """
+    Return the first COUNT objects passed in the pipeline
+
+    EXAMPLES
+        Print the name of the first pool
+
+            sdb> spa | head 1 | spa_name
```
sdimitro: I think the command here would be `spa | head 1 | member spa_name` (the word `member` is missing above)

ctuffli:
OK, I've attempted to rebase my changes on the latest and think I'm close, but seem to be hitting a failure in pytest. Any pointers as to how I need to fix the test failures?

sdimitro:
Seems to be a problem with new changes introduced in `drgn` and some misconfiguration in `libkdumpfile` - I'll need to investigate

ahrens:
@ctuffli It will be great to get this integrated.  It looks like there are a couple of outstanding comments from Serapheim.  I just wanted to remind you to take a look at them when you get a chance.

ctuffli:
@ahrens most of @sdimitro comments should be addressed. I've rebased my branch off of the latest delphix/sdb and pushed.

sdb/commands/pretty_print.py
```diff
@@ -23,6 +23,9 @@
 
 
 class PrettyPrint(sdb.Command):
+    """
+    TODO
+    """
```
ahrens: Is it useful to have this here, as opposed to not having a specific help message?

sdb/commands/pyfilter.py
```diff
@@ -24,6 +24,9 @@
 
 
 class PyFilter(sdb.Command):
+    """
+    TODO
+    """
```
ahrens: same here

sdb/commands/stacks.py
ahrens: Are you planning to add these examples in this PR?

sdb/commands/pyfilter.py
ctuffli: I wasn't clear what the difference was between filter and pyfilter. Would the following work? Other examples you use on a regular basis?
```
Return objects matching the Python expression

EXAMPLES
    Print the even addresses

        sdb> echo 0 1 2 3 4 | cast int | pyfilter obj % 2 == 0
        (int)0
        (int)2
        (int)4
```

sdb/commands/pretty_print.py
ctuffli: Sorry, this was meant as a indication I need help :)
It appears that most things do a 'pretty print' by default. When would you need to call this command directly?

sdb/commands/echo.py
```diff
@@ -24,6 +24,9 @@
 
 
 class Echo(sdb.Command):
+    """
+    Creates an object of type (void *) from the give address or addresses
```
prakashsurya: "give" -> "given"

sdb/commands/stacks.py
```diff
@@ -52,6 +52,78 @@
...
+    Print the stack traces for the active tasks / threads
+
+    By default, the command will aggregate similar call stacks
+    printing them in descending order of frequency. The output
+    includes the `struct task_struct` address, thread state, and
+    aggregation count.
+
+    Optionally, the command can filter stacks, displaying only
+    those that match a given thread state, containing a given
+    function, or belonging to a given kernel module.
```
prakashsurya: We should add another level of indentation here, so it aligns better when printed via the `help` command.

sdb/commands/pyfilter.py
prakashsurya: I'd recommend we just leave this undocumented for now.

In general, I think folks should be using `filter`.. but I think there are more complex cases where `filter` won't work, which is when `pyfilter` could be used.. since the cases where `pyfilter` is required should be less common, I don't think it's critical to have this command documented (for now, at least).

sdb/commands/pretty_print.py
prakashsurya: I'd be OK leaving this undocumented for now, since as you say, most things will pretty print by default.

sdb/commands/echo.py
ctuffli: Fixed

sdb/commands/stacks.py
ctuffli: Fixed

sdb/commands/pretty_print.py
ctuffli: Removed

sdb/commands/pyfilter.py
ctuffli: Removed

PR #113, opened 2019-12-05
Introduce mypy to the codebase

sdimitro:


sdb/commands/zfs/btree.py
```diff
@@ -52,6 +52,8 @@ class Btree(sdb.Walker):
     names = ["zfs_btree"]
     input_type = "zfs_btree_t *"
 
+    elem_size: drgn.Object
```
prakashsurya: Is this describing a class level variable, or an instance variable? Presumably this is for the instance variable, but it strikes me as odd, since this is where class variables would be defined (e.g. the two above).

prakashsurya:
If we want to ensure the mypy test is "required" when merging PRs (like the other tests), we'll have to get a repository admin to configure this for us (e.g. perhaps @sebroy can do this?).

sdb/commands/zfs/btree.py
sdimitro: good catch - should be fixed now.
Take another look at the change and let me know if you find anything else otherwise let's merge this.

PR #117, opened 2019-12-10
Upload test coverage reports to codecov.io

prakashsurya:


prakashsurya:
As far as I understand, the codecov token required to upload the coverage report is not available to PRs, since we're using github secrets to expose the token to the github action. Thus, we can't really test this via the PR; but the token should be available to the action once this lands to the master branch.

This does mean, though, that we can't leverage the PR code coverage status functionality of codecov, which is unfortunate.

PR #121, opened 2020-01-09
Want lxlist command

### LINKED ISSUE #25: Want `lx_list` command
Most SDB devs/users have implemented some variant of this command on their own. Let's standardize it.

Looking at my home directory I have the following:
```
class LxList(sdb.Command):
    # pylint: disable=too-few-public-methods

    names = ["lxlist"]

    # t o d o: validate 'type' and 'member' passed in, and handle errors nicely
    # for instance, passing "asdf for a type will complain with a SyntaxError ?

    def _init_argparse(self, parser: argparse.ArgumentParser) -> None:
        parser.add_argument(
            "struct_name",
            help="name of the struct used for entries in the list")
        parser.add_argument(
            "member", help="name of the hlist_node member within the struct")

    def __get_valid_struct_name(self, name: str) -> str:
        #
        # HACK: Since a proper lexer is still not in place we pass
        # structure names without the 'struct' keyword before
        # them. Thus, we first check if the type passed is a typedef
        # and if we don't find any type with that name we add
        # the 'struct' keyword ourselves and look again.
        #
        try:
            type_ = self.prog.type(name)
        except LookupError:
            type_ = None

        # Found typedef - return
        if type_:
            return type_

        # Check for 'struct'
        struct_prefixed = "struct {:s}".format(name)
        try:
            type_ = self.prog.type(struct_prefixed)
        except LookupError as err:
            raise sdb.CommandError(self.name, str(err))
        return struct_prefixed

    def call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
        name = self.__get_valid_struct_name(self.args.struct_name)
        #
        # XXX: It would be nice once we have the actual type here
        # to validate structs (if it's a typedef we'd need to resolve it
        # first) and ensure that the list member used for the iteration
        # is valid.
        #
        for obj in objs:
            yield from list_for_each_entry(name, obj, self.args.member)
```

The usage of it is/would be something along those lines:
```
> addr slab_root_caches | lxlist kmem_cache memcg_params.__root_caches_node | cast void *
(void *)0xffff8c1ceda40f00
(void *)0xffff8c1ceda40780
(void *)0xffff8c1ceda41c80
(void *)0xffff8c1ceda41380
(void *)0xffff8c1cc0626600
...
```

sdimitro:
= Patch

This patch implements sdb commands to walk `list_head`s and
`hlist_head`s as provided by the Linux kernel.

Closes #25


sdimitro:
cc @jgallag88

PR #124, opened 2020-01-09
Want container_of command

sdimitro:


sdimitro:
cc @jgallag88

sdb/commands/container_of.py
```diff
@@ -0,0 +1,65 @@
...
+import sdb
+from sdb.commands.internal.util import get_valid_struct_name
+
+
+class ContainerOf(sdb.Command):
+    """
+    Get the containing object of a pointer object.
+
+    EXAMPLES
+        Superfluous example of getting a structure's address
```
ahrens: `Superfluous` -> `Trivial` ?  If it's really superfluous then we can remove it :)

sdb/commands/internal/util.py
```diff
@@ -0,0 +1,66 @@
...
+# pylint: disable=missing-docstring
+
+import drgn
+import sdb
+
+
+def get_valid_struct_name(cmd: sdb.Command, tname: str) -> str:
+    """
+    If tname is a name of a type that's a typedef to a
+    struct, the function return it as is. If this is a
```
ahrens: `the function return` -> `this function will return`?

sdimitro:
@prakashsurya @jgallag88 can I get a second review on this as Matt's became stale after I applied his feedback.

sdb/commands/container_of.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Get the containing object of a pointer object.
+
+    EXAMPLES
+        Trivial example of getting a structure's address
+        out of the address of one of its members:
+
+            sdb> addr init_task | cast void *
+            (void *)0xffffffffa8217740
+            sdb> addr init_task | member comm | addr | container_of struct task_struct comm  | cast void *
```
prakashsurya: I don't mean for this to be a blocking comment.. but I'm curious.. do we always have to provide the structure and the member?

e.g. if we did this instead:
```
addr init_task | member comm | container_of  | cast void *
```
could this work?

I'm wondering, since we probably could (not sure if we actually do) have all the type information passed across the pipes..

e.g. `addr init_task` should know the type of `init_task` is `struct task_struct`.. and then `member` knows it's pulling a member from a `struct task_struct`.. so then, perhaps `container_of` could know that it was passed the `comm` member of the `struct task_struct` structure, just from the information in the pipe, and automatically do the right thing..

I don't know if all of the necessary information is passed into `container_of` right now, but I thought I'd at least ask.. and if not, would it be useful to try and achieve something like this (i.e. open a bug, and leave it for later work)?

ahrens: `container_of` only get the object that's passed to it, which doesn't know what its "container" is.  We could theoretically make `member` pass this information through somehow, but this is a contrived use.  If you have the container already you wouldn't use `|member ... | container_of`.  Typically you would have the struct pointer from some other source and happen to know that it's inside of another struct.  E.g. I have an `avl_node_t*` which I happen to know is `dmu_buf_impl_t`'s `db_link` and I want to get the containing `dmu_buf_impl_t*`.

So I don't think it's practical for `container_of` to figure out what the containing type is.

prakashsurya: :+1: yea, ok.. I think that makes sense.

PR #127, opened 2020-01-14
Want array command

sdimitro:


sdimitro:
cc @jgallag88

sdb/commands/array.py
```diff
@@ -0,0 +1,133 @@
...
+
+            sdb> spa | member spa_zio_taskq[0][0].stqs_taskq | array 2
+            *(taskq_t *)0xffff90cc127ea600 = {
+                ... <cropped> ...
+            }
+            *(taskq_t *)0xffff90cc181bfb00 = {
+                ... <cropped> ...
+            }
+
+        Until a better way/notation is implemented on SDB the only
```
ahrens: Out of curiosity, do we need infrastructure work to do this, or is it a matter of adding parsing to the array command to handle something like `array 3-5`?

sdb/commands/array.py
```diff
@@ -0,0 +1,133 @@
...
+            nelems = 0
+
+            obj_type = sdb.type_canonicalize(obj.type_)
+            if obj_type.kind == drgn.TypeKind.ARRAY:
+                array_elems = len(obj)
+                if self.args.nelems is not None:
+                    nelems = self.args.nelems
+                    if nelems > array_elems:
+                        print(
+                            f"warning: exceeding array size {array_elems} - requested {nelems}"
```
ahrens: [nit] How about `warning: requested size {nelems} exceeds type size {array_elems}`

sdb/commands/array.py
```diff
@@ -0,0 +1,133 @@
...
+                        )
+                else:
+                    nelems = array_elems
+
+            elif obj_type.kind == drgn.TypeKind.POINTER:
+                if self.args.nelems is None:
+                    err_msg = f"'{obj.type_.type_name()}' is a pointer - "
+                    err_msg += "please specify the number of elements"
```
ahrens: You can use implicit concatenation:
```
err_msg = f"..."
  f"..."
```

sdb/commands/array.py
```diff
@@ -0,0 +1,133 @@
...
+                    raise sdb.CommandError(self.name, err_msg)
+                nelems = self.args.nelems
+
+                if not obj_type.type.is_complete():
+                    err_msg = "can't walk pointer array of incomplete type "
+                    err_msg += f"'{obj_type.type.type_name()}'"
```
ahrens: same here

sdb/commands/array.py
```diff
@@ -0,0 +1,133 @@
...
+                # pointer array first to ensure that it is
+                # valid before attempting to dereference it
+                # later
+                #
+                try:
+                    obj[0].read_()
+                except drgn.FaultError:
+                    raise sdb.CommandError(
+                        self.name,
+                        f"invalid memory access at address {hex(obj.value_())}")
```
ahrens: I wonder if we should be catching this at a higher level, in the sdb infrastructure.  You could have other kinds of memory errors, like going off the end of the array into unmapped memory.

sdb/commands/array.py
sdimitro: I don't think we need to do anything in the infrastructure. We should be able to bypass argparse (e.g. just use it to give us the leftover tokens that it doesn't understand) and use our own parsing logic.

sdimitro:
With https://github.com/delphix/sdb/commit/22acd08e36b40fd22be86d6f1c6373496a9e0c0b commit in, I updated this and it should be ready for another round.

sdb/commands/array.py
```diff
@@ -0,0 +1,119 @@
...
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        parser = super()._init_parser(name)
+        parser.add_argument("nelems",
+                            type=int,
+                            nargs="?",
+                            help="number of elements in array to walk")
+        return parser
+
+    def _call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
+        for obj in objs:
```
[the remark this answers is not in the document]
sdimitro: Good idea! Done!

PR #130, opened 2020-01-14
Fix dereferencing bad addresses in member command

### LINKED ISSUE #125: internal error when "member" dereferences bad pointer
```
sdb> echo 0x1234 | cast dmu_recv_cookie_t * | member drc_os
sdb encountered an internal error due to a bug. Here's the
information you need to file the bug:
----------------------------------------------------------
Target Info:
	ProgramFlags.IS_LINUX_KERNEL
	Platform(<Architecture.X86_64: 1>, <PlatformFlags.IS_LITTLE_ENDIAN|IS_64_BIT: 3>)

  File "/usr/lib/python3/dist-packages/sdb/internal/repl.py", line 86, in eval_cmd
    print(obj)
_drgn.FaultError: kdump_read: Cannot get page I/O address: PDPT table not present: pgd[0] = 0x0
----------------------------------------------------------
Link: https://github.com/delphix/sdb/issues/new
```

Note this is basically the same as #8, but that was closed when NULL pointer handling was fixed.  But invalid-but-not-NULL pointers still cause this problem.

sdimitro:
Add integration tests for member command
Cleanup member unit tests
Factor out helper for checking valid memory deref

Closes #125


sdb/commands/member.py
```diff
@@ -289,16 +289,17 @@ def _eval_member_terms(
             self._validate_type_dereference(obj, sep)
 
             #
-            # Ensure we are not trying to dereference NULL.
+            # Ensure we are not trying to dereference NULL or
+            # any kind of invalid memory.
             #
-            if obj.value_() == 0x0:
-                warning = f"cannot dereference NULL member of type '{obj.type_}'"
+            if not sdb.is_valid_memory_access(obj):
```
ahrens: I wonder if we should be letting drgn throw the memoy access exception and then catching it in the sdb infrastructure, outside of specific commands?

sdimitro: I just issued https://github.com/delphix/sdb/issues/131 .

sdimitro:
Closing in favor of https://github.com/delphix/sdb/pull/134

PR #134, opened 2020-01-16
Implement _call_one() method for Commands

sdimitro:
Handle all uncaught invalid memory accesses in the pipeline
Migrate positive tests cases of filter and member to integration tests
Fix new pylint error for input_type

Note #1:
Test cases from filter and member had to be migrated because some
of our mocks were not backed by valid memory and the pipeline can
now detect that, that's why this change was bundled together. I
decided to not let go of the negative tests yet though as they
are still useful in checking that the correct programmatic
exception is thrown.

Note #2:
The error messages introduced when dereferencing invalid memory
are not optimal, partly because drgn doesn't give us a lot of
flexibility. The plan is to re-write them properly once we have
the relevant infrastructure ready underneath us.


sdb/command.py
```diff
@@ -225,14 +234,145 @@ def __init_subclass__(cls, **kwargs):
 
     def _call(self,
               objs: Iterable[drgn.Object]) -> Optional[Iterable[drgn.Object]]:
-        # pylint: disable=missing-docstring
-        raise NotImplementedError
+        """
+        Optionally implemented by the user.
```
ahrens: user -> subclass?

sdb/command.py
```diff
@@ -225,14 +234,145 @@ def __init_subclass__(cls, **kwargs):
...
+
+        Commands that expect to manipulate and deal with any errors
+        of all the objects from the previous stage of the pipeline
+        passed as an iterable, should implement this method.
+
+        Takes precedence over _call_one() - see call() method.
```
ahrens: should we say, subclasses should override exactly one of _call or _call_one (not both)?

sdb/command.py
```diff
@@ -225,14 +234,145 @@ def __init_subclass__(cls, **kwargs):
...
     def _call(self,
               objs: Iterable[drgn.Object]) -> Optional[Iterable[drgn.Object]]:
-        # pylint: disable=missing-docstring
-        raise NotImplementedError
+        """
+        Optionally implemented by the user.
+
+        Commands that expect to manipulate and deal with any errors
+        of all the objects from the previous stage of the pipeline
+        passed as an iterable, should implement this method.
```
ahrens: I think it's not just error handling - e.g. they might want to maintain state between each object (to count them or summarize them in some way)

sdb/command.py
```diff
@@ -225,14 +234,145 @@ def __init_subclass__(cls, **kwargs):
...
+        Takes precedence over _call_one() - see call() method.
+        """
+        #
+        # We should never get here!
+        #
+        assert self.name is None
+        assert False
+        return objs
+
+    def _call_one(self, obj: drgn.Object) -> Optional[drgn.Object]:
```
ahrens: Maybe this should return `Iterable[drgn.Object]`, so that it could e.g. find and yield multiple objects inside the `obj`.  That said, there are not many _call methods that yield multiple things per input object (AFAIK only the abstract classes Walker and Locator do).  In fact there are not many _call methods period, since most commands are PrettyPrinter and/or Locators.

sdb/command.py
```diff
@@ -225,14 +234,145 @@ def __init_subclass__(cls, **kwargs):
...
+        """
+        Optionally implemented by the user.
+
```
ahrens: I wonder if a better mechanism for the "call one" would be to have an abstract subclass  of Command (e.g. `SingleInputCommand`) that would do the iteration and print/discard the exception?  And then if you want to use this, you would subclass from SingleInputCommand and override its _call_one method.

sdimitro: So I did think about this as it would probably help us with some of the trickiness of the current solution, but then I thought about the scenario of someone trying to create a command that is a `SingleInputCommand, Locator, PrettyPrinter`. Things should work as expected maybe(?) but I couldn't tell if that'd be more tricky (weird inheritance hierarchy) than what we already have. I can create a separate branch maybe an we can compare the two.

sdb/command.py
sdimitro: That was a great catch! `member` actually can return multiple objects already from one input object - when you specify the dereference of multiple fields of that struct:
```
--- a/tests/integration/data/regression_output/core/addr spa_namespace_avl | member avl_root.avl_pcb avl_size
+++ b/tests/integration/data/regression_output/core/addr spa_namespace_avl | member avl_root.avl_pcb avl_size
@@ -1 +1,2 @@
 (uintptr_t)1
+(size_t)9176
```

sdimitro:
Closed in favor of https://github.com/delphix/sdb/pull/135

PR #135, opened 2020-01-17
Implement _call_one() method for Commands

sdimitro:
Handle all uncaught invalid memory accesses in the pipeline
Migrate positive tests cases of filter and member to integration tests
Fix new pylint error for input_type

Note #1:
Test cases from filter and member had to be migrated because some
of our mocks were not backed by valid memory and the pipeline can
now detect that, that's why this change was bundled together. I
decided to not let go of the negative tests yet though as they
are still useful in checking that the correct programmatic
exception is thrown.

Note #2:
The error messages introduced when dereferencing invalid memory
are not optimal, partly because drgn doesn't give us a lot of
flexibility. The plan is to re-write them properly once we have
the relevant infrastructure ready underneath us.


sdb/command.py
```diff
@@ -225,14 +234,104 @@ def __init_subclass__(cls, **kwargs):
...
+                    result, not issubclass(self.__class__, SingleInputCommand))
+        except drgn.FaultError as err:
+            raise CommandError(self.name, f"invalid memory access: {str(err)}")
+
+
+class SingleInputCommand(Command):
+    """
+    Commands that are expected to recover from bad memory accesses
+    implicitly by the sdb infrastructure and move on to the next
+    object should subclass this class.
```
ahrens: I think the bad memory access recovery is nice but maybe we can put the principle of processing each object independently up front.  I could imagine wanting to recover from additional errors in the future (e.g. ArrayBoundsError, once we do that):
`Commands that would like to process each input object independently (without saving state between objects) can subclass from this class.  If a FaultError exception is thrown while processing one object, processing will continue with the next object.`

sdb/command.py
```diff
@@ -493,6 +577,9 @@ class Walker(Command):
     """
     A walker is a command that is designed to iterate over data
     structures that contain arbitrary data types.
+
+    Note: Currently, a Walker cannot be defined to also be a
+          SingleInputCommand.
```
ahrens: I think it might go without saying that multiple inheritance (especially from multiple abstract classes of a single base class) is tricky and being able to do it would be the exception rather than the norm (e.g. inheriting from both PrettyPrinter and Locator, which requires special code in `Locator._call()`, and knowing the order you should inherit from them in).

Put another way, you can't be a Walker, PrettyPrinter, or Locator, and also define your own _call method.  Likewise you can't expect to define your own _call_one method.

I think it's sufficient to explain this in the comment for SingleInputCommand.

sdb/command.py
```diff
@@ -225,14 +234,104 @@ def __init_subclass__(cls, **kwargs):
...
+                # SingleInputCommand and we set the `fatal` flag
+                # accordinly.
+                #
+                yield from self.__invalid_memory_objects_check(
+                    result, not issubclass(self.__class__, SingleInputCommand))
+        except drgn.FaultError as err:
+            raise CommandError(self.name, f"invalid memory access: {str(err)}")
+
+
+class SingleInputCommand(Command):
```
ahrens: I think we can change Type, Filter, and RangeTreeSeg, to subclass from this as well.

It's a testament to our design of the PrettyPrinter / Locator / Walker paradigm that there are not more plain Commands that can be converted to SingleInput Command!  The plain Commands (and SingleInputCommands) are the exception.

sdimitro: Good point! Done!

PR #139, opened 2020-01-23
Want print command for rudimentary formatting of output

sdimitro:
Default to printing just address for pointer objects

PR Note:
The motivation for this was to help with use-cases where sdb output is captured in a shell variable for later processing/use (e.g https://github.com/delphix/zfs/pull/125)


sdimitro:
cc @jwk404

sdb/commands/print.py
```diff
@@ -0,0 +1,100 @@
...
+
+    EXAMPLES
+        By default getting the address of a global variable
+        looks like this:
+
+            sdb> addr spa_namespace_avl
+            (avl_tree_t *)spa_namespace_avl+0x0 = 0xffffffffc0954fe0
+
+        One could use the `print` command to format the above
+        to something that is better for consumption like this:
```
ahrens: `for programmatic consumption` ?

sdb/commands/print.py
```diff
@@ -0,0 +1,100 @@
...
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        #
+        # Check the following for available options that are built-in to drgn:
+        # drgn.readthedocs.io/en/latest/api_reference.html#drgn.Object.format_
+        #
+        parser = super()._init_parser(name)
+        parser.add_argument("-c",
+                            "--char",
+                            action="store_true",
+                            default=False,
```
ahrens: I think that `default` may not be needed here, since according to the docs:
> 'store_true' and 'store_false' - These are special cases of 'store_const' used for storing the values True and False respectively. In addition, they create default values of False and True respectively.

sdb/commands/print.py
```diff
@@ -0,0 +1,100 @@
...
+            action="store_true",
+            default=False,
+            help=("don't symbolize and don't print type names,"
+                  " member names in structs, nor indices in arrays"))
+        return parser
+
+    def _call_one(self, obj: drgn.Object) -> None:
+        if self.args.RAW:
+            self.args.raw = True
+            self.args.nosymbolize = True
```
ahrens: [nit] It seems a little unconventional to be changing `self.args.*` for each object.  The simplest way around this would be to have local variables, like `raw = self.args.raw or self.args.RAW`.

Another option would be to use a custom Action subclass that would store True in all 3 members.  See https://docs.python.org/3/library/argparse.html#argparse.Action

sdb/commands/print.py
sdimitro: You are right! Thanks for pointing that out

sdb/commands/print.py
sdimitro: I updated this and rebased with the latest array changes - let me know what you think

PR #146, opened 2020-02-05
Add a few helpers for quickly finding task_structs and file structs within them

sdimitro:
There are 2 commits here. The first one is a side-fix for an f-string bug (see changes in sdb/error.py).

### DIFF
--- a/sdb/commands/linux/process.py
+++ b/sdb/commands/linux/process.py
@@ -1,0 +1,90 @@
+#
+# Copyright 2020 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pylint: disable=missing-docstring
+
+import argparse
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class FindPid(sdb.Command):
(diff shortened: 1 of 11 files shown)


sdimitro:
cc @grwilson

sdb/commands/linux/process.py
```diff
@@ -0,0 +1,90 @@
...
+            .count = (refcount_t){
+                    .refs = (atomic_t){
+                            .counter = (int)50,
+                    },
+            },
+            .level = (unsigned int)0,
+            ...
+    """
+
+    names = ["find_pid"]
```
ahrens: why "find_pid" rather than plain "pid"?

sdb/commands/linux/process.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @classmethod
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        parser = super()._init_parser(name)
+        parser.add_argument("pid",
+                            type=int,
+                            nargs="+",
+                            help="numeric PID(s) of process(es)")
```
ahrens: Seems more usable than `echo 4 | pid` 👍

sdb/commands/linux/vfs.py
```diff
@@ -0,0 +1,51 @@
...
+import argparse
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class FGet(sdb.SingleInputCommand):
+    """
+    Return "struct file *" given the FD number for each task_struct piped as input.
```
ahrens: I think if you declare the input_type, we get some benefits like auto casting and type checking.  unless you are intentionally relying on `drgn.helpers.linux.fs.fget()` for that?

sdb/commands/linux/process.py
sdimitro: Changed

sdb/commands/linux/vfs.py
sdimitro: No I wasn't. Thanks! input_type added

PR #147, opened 2020-02-07
Introduce ptype command

sdimitro:
Note that the argument parsing could be made smarter (and more
complex) to allow the user to use C keywords like "struct"
before the type. We just leave it to SDB to figure out by
itself for now until we figure out how we want to deal with this
in this and other commands in the repo.


sdimitro:
cc @jgallag88

sdb/commands/ptype.py
```diff
@@ -0,0 +1,92 @@
...
+import argparse
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class Type(sdb.Command):
+    """
+    Print the given type passed as an argument.
```
jgallag88: `nit`: 'given type passed as an argument' is redundant, maybe drop 'given'?

sdb/commands/ptype.py
```diff
@@ -0,0 +1,92 @@
...
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        parser = super()._init_parser(name)
+        parser.add_argument("type", nargs="*", metavar="<type>")
+        return parser
+
+    def _call(self, objs: Iterable[drgn.Object]) -> None:
+        for tp_name in self.args.type:
+            found = False
+
+            for prefix in ["struct ", "enum ", "union ", ""]:
```
ahrens: [nit] we might want to check the exact name they said first (i.e. list `""` first)

sdimitro: I just realized that I should have created a comment about this but I'd like to know your opinion.

There are cases in C codebases that someone may do this:
```
typedef union GCObject GCObject;
```
[example from `module/lua/lobject.h` in the ZFS repo (actually Lua repo)]

This is kind of annoying because using `ptype` if we have `""` before all the other strings in that list we get:
```
sdb> ptype GCObject
typedef union GCObject GCObject
```

If we have the code as I have it though, typedef's are the last thing that we check:
```
sdb> ptype GCObject
union GCObject {
	GCheader gch;
	union TString ts;
	union Udata u;
	union Closure cl;
	struct Table h;
	struct Proto p;
	struct UpVal uv;
	struct lua_State th;
}
```

It is an interesting issue which would have been nice if we supported things like `ptype union GCObject` and `ptype GCObject`.

sdimitro: Comment added.

PR #148, opened 2020-02-07
Introduce per-CPU commands

sdimitro:


sdimitro:
cc @jgallag88 @pzakha

sdb/commands/linux/per_cpu.py
```diff
@@ -0,0 +1,92 @@
...
+# pylint: disable=missing-docstring
+
+import argparse
+from typing import Iterable
+
+import drgn
+import sdb
+
+
+class LxPerCpuPtr(sdb.SingleInputCommand):
```
ahrens: can we declare the input_type (for both commands)?

sdb/commands/linux/per_cpu.py
```diff
@@ -0,0 +1,93 @@
...
+
+    EXAMPLES
+        Get the per-CPU data of the "kmalloc-8" kmem cache for CPUs 0 and 1:
+
+            sdb> slabs | filter obj.name == "kmalloc-8" | member cpu_slab
+            (struct kmem_cache_cpu *)0x2f040
+            sdb> slabs | filter obj.name == "kmalloc-8" | member cpu_slab | cpu_ptr 0
+            (struct kmem_cache_cpu *)0xffff944effc2f040
+            sdb> slabs | filter obj.name == "kmalloc-8" | member cpu_slab | cpu_ptr 1
+            (struct kmem_cache_cpu *)0xffff944effd2f040
```
jgallag88: What do you think about allowing the user to omit the cpu number, in which case the command will emit the pointer for each CPU that exists? That feels like an idiomatic way for the command to work. Also, I suspect that people will usually be less interested in the values on particular CPUs, and more interested in aggregating the values on all of the CPUs. Emitting a list of values would make it simple to do all sorts of aggregations.

sdb/commands/linux/per_cpu.py
sdimitro: I added the input type for the 2nd command as it expects a specific struct.
`LxPerCpuPtr` is different as it literally accepts ANY pointer type. If we wanted to be EXTRA in sdb (the Linux Kernel I think doesn't actually break that way) is to ensure that pointer type passed also has the `__percpu` type specifier in its type info. Do you think that'd be useful?

sdb/commands/linux/per_cpu.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    names = ["cpu_ptr"]
```
jgallag88: Since linux puts "percpu" in the name of everything that relates to these variables, I think it might make sense to include that in the name of the command as well, to make it easier to guess what the command does. If we went as far as removing the ability to specify which CPU, and just always output the pointers for all the CPUs (as suggested above), we could potentially just call the command `percpu`.

sdb/commands/linux/per_cpu.py
sdimitro: After discussing this with @jgallag88 , we decided to check for this and print out a warning if the value is not `__percpu`.

sdimitro: @jgallag88 I should have checked the drgn documentation first and the kernel but apparently this is not exactly a qualifier so we can't figure this out from drgn. As a result I won't be able to make any extra tests.

sdb/commands/linux/per_cpu.py
sdimitro: Changed it so without an arguments we print all CPUs but still allowing the user to specify an a specific CPU or set of CPUs if they so wish.

sdb/commands/linux/per_cpu.py
sdimitro: Name changed. Thank you!

tests/integration/data/regression_output/linux/slabs | filter obj.name == "kmalloc-8" | member cpu_slab | percpu 0 2 1
```diff
@@ -0,0 +1,2 @@
+(struct kmem_cache_cpu *)0xffffa0897fc27040
+sdb: percpu: available CPUs [0-1] - requested CPU 2
```
pzakha: How is regression testing performed? Do we know for sure that the test environment will have 2 CPUs?

sdimitro: This is ran against a reference crash dump that we have and we know that it will always have 2 CPUs on it.

pzakha: Ah okay, thanks for the info

Issue #160, opened 2020-02-13
libkdumpfile build failures in automated testing


sdimitro:
```
/bin/bash ../libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I..  -I../include  -I/opt/hostedtoolcache/Python/3.6.10/x64/include/python3.6m -I/opt/hostedtoolcache/Python/3.6.10/x64/include/python3.6m  -Wno-unused-result -Wsign-compare  -DNDEBUG -g -fwrapv -O3 -Wall -g -O2 -MT kdumpfile.lo -MD -MP -MF .deps/kdumpfile.Tpo -c -o kdumpfile.lo kdumpfile.c
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I.. -I../include -I/opt/hostedtoolcache/Python/3.6.10/x64/include/python3.6m -I/opt/hostedtoolcache/Python/3.6.10/x64/include/python3.6m -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -g -O2 -MT kdumpfile.lo -MD -MP -MF .deps/kdumpfile.Tpo -c kdumpfile.c  -fPIC -DPIC -o .libs/kdumpfile.o
kdumpfile.c:2045:8: error: ‘Py_TPFLAGS_HAVE_NEWBUFFER’ undeclared here (not in a function); did you mean ‘Py_TPFLAGS_HAVE_FINALIZE’?
      | Py_TPFLAGS_HAVE_NEWBUFFER, /* tp_flags */
Makefile:776: recipe for target 'kdumpfile.lo' failed
        ^~~~~~~~~~~~~~~~~~~~~~~~~
make[3]: Leaving directory '/home/runner/work/sdb/sdb/libkdumpfile/python'
        Py_TPFLAGS_HAVE_FINALIZE
make[3]: *** [kdumpfile.lo] Error 1
Makefile:795: recipe for target 'all-recursive' failed
make[2]: *** [all-recursive] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
make[2]: Leaving directory '/home/runner/work/sdb/sdb/libkdumpfile/python'
Makefile:521: recipe for target 'all-recursive' failed
make[1]: Leaving directory '/home/runner/work/sdb/sdb/libkdumpfile'
Makefile:428: recipe for target 'all' failed
##[error]Process completed with exit code 2.
```

[...]

PR #162, opened 2020-02-14
Disable python bindings for libkdumpfile temporarily

sdimitro:
Due to issue #160 introduced by recent changes in libkdumpfile
our nightly and PR automation have been failing because the
library can't compile. Until the issue is resolved in the library
itself or on our side (by creating a proper environment for the
github action) we disable the compilation of the Python bindings
of the library. The impact is almost non-existent as these
bindings are not used in any way by SDB nor DRGN.


sdimitro:
Petr jumped on this right away with ptesarik/libkdumpfile@2e8cec4 so this issue now should be fixed. I'll close it once I verify the fix.

Issue #154, opened 2020-02-11
Want a way to find a stack for a given thread


shartse:
Either something like `::findstack`, or using stacks as a location which accepts a `task_stuct`, or add a flag to `threads`.

PR #164, opened 2020-02-14
Stacks should be a pretty printer and a locator

shartse:
## Pull request checklist

- [X] Build was run locally and any changes were pushed
- [X] Lint has passed locally and any fixes were made for failures

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

resolves https://github.com/delphix/sdb/issues/154

## Decription
Currently, `stacks` is a stand-alone command that prints information about all the stacks of threads on the system. As per https://github.com/delphix/sdb/issues/154, we wanted a way to be able to display the stack for a specific thread (`task_struct`), so it makes sense to have `stacks` be a `PrettyPrinter`. Now we can do something like this:

```
sdb> echo 0xffff9dad75fadb00 | stack
0xffff9dad75fadb00 INTERRUPTIBLE         1
                  __schedule+0x24e
                  schedule+0x2c
                  oom_reaper+0x152
                  kthread+0x121
                  ret_from_fork+0x1f
```
However, we also want to maintain the original "print all the stacks" behavior, which indicates that `stacks` should be a `Locator` as well. I refactored the code so that the filter arguments (by state, function, and module) also apply to the `Locator` behavior, which means we can now do something like this:

```
sdb> stacks | wc
(unsigned long long)459
sdb> stacks -m zfs | wc
(unsigned long long)19
```
In many ways, `stacks` is a more specialized version of `threads`, so I think the most useful thing is to just have it pass along `task_stuct` objects as it's output value. 

- [ ] Yes
- [X] No

## Testing
This re-factor does not change the initial behavior of the `stacks` command (I was able to verify this by seeing that the previous tests still pass)

### DIFF
--- a/sdb/commands/stacks.py
+++ b/sdb/commands/stacks.py
@@ -17,7 +17,7 @@
 # pylint: disable=missing-docstring
 
 import argparse
-from typing import Any, Dict, Iterable, List, Tuple
+from typing import Dict, Iterable, List, Tuple
 from collections import defaultdict
 
 import drgn
@@ -27,34 +27,11 @@ from drgn.helpers.linux.pid import for_each_task
 import sdb
 
 
-#
-# Note: This is a rudimentary version of what the command could/should be.
-#
-# On the high-level, it could be a `Locator`, or something similar, where
-# objects could be passed as input from the pipeline dispatching different
-# methods depending on the type of object. E.g. input could be a namespace
-# object and we print all the task structs within it, or it could be just
-# a list of task structs passed from a previous command for filtering.
-# Another option would be to decouple the stack listing, filtering, and
-# pretty-printing functionality to independent SDB commands.
-#
-# There are also other lower-level usability improvements like supporting
-# filtering by `function+offset` with the `-c` option, or by namespace ID
-# using `-n <ID>`.
-#
-# Finally, the command lacks any support for userland targets.
-#
-# SDB is still in its early stages and hasn't been used enough for us to
-# be clear which use cases really matter. In the meantime if we don't have
-# anything that provides this functionality it won't be easy to do this
-# exploration. The version below is a good enough for the time being
-# providing some basic functionality and being our tracer bullet for
-# future iterations.
-#
-class Stacks(sdb.Command):
+class Stacks(sdb.Locator, sdb.PrettyPrinter):
     """
-    Print the stack traces for the active tasks / threads
+    Print the stack traces for active threads (task_struct)
 
+    DESCRIPTION
         By default, the command will aggregate similar call stacks
         printing them in descending order of frequency. The output
         includes the `struct task_struct` address, thread state, and
@@ -64,8 +41,11 @@ class Stacks(sdb.Command):
         those that match a given thread state, containing a given
         function, or belonging to a given kernel module.
 
-        EXAMPLES
-            Print the call stacks for all tasks
+        The command returns all task_stuct structs that matched the
+        filter.
+
+    EXAMPLES
+        Print the call stacks for all tasks
 
             sdb> stacks
             TASK_STRUCT        STATE             COUNT
@@ -85,47 +65,89 @@ class Stacks(sdb.Command):
                               ret_from_fork+0x35
             ...
 
-            Print stacks containing functions from the zfs module
+        Print stacks containing functions from the zfs module
 
-                sdb> stacks -m zfs
-                TASK_STRUCT        STATE             COUNT
-                ==========================================
-                0xffff952130515940 INTERRUPTIBLE         1
-                                  __schedule+0x24e
-                                  schedule+0x2c
-                                  cv_wait_common+0x11f
-                                  __cv_wait_sig+0x15
-                                  zthr_procedure+0x51
-                                  thread_generic_wrapper+0x74
-                                  kthread+0x121
-                                  ret_from_fork+0x35
-                ...
+            sdb> stacks -m zfs
+            TASK_STRUCT        STATE             COUNT
+            ==========================================
+            0xffff952130515940 INTERRUPTIBLE         1
+                              __schedule+0x24e
+                              schedule+0x2c
+                              cv_wait_common+0x11f
+                              __cv_wait_sig+0x15
+                              zthr_procedure+0x51
+                              thread_generic_wrapper+0x74
+                              kthread+0x121
+                              ret_from_fork+0x35
+            ...
 
-            Print stacks containing the l2arc_feed_thread function
+        Print stacks containing the l2arc_feed_thread function
 
-                sdb> stacks -c l2arc_feed_thread
-                TASK_STRUCT        STATE             COUNT
-                ==========================================
-                0xffff9521b3f43b80 INTERRUPTIBLE         1
-                                  __schedule+0x24e
-                                  schedule+0x2c
-                                  schedule_timeout+0x15d
-                                  __cv_timedwait_common+0xdf
-                                  __cv_timedwait_sig+0x16
-                                  l2arc_feed_thread+0x66
-                                  thread_generic_wrapper+0x74
-                                  kthread+0x121
-                                  ret_from_fork+0x35
+            sdb> stacks -c l2arc_feed_thread
+            TASK_STRUCT        STATE             COUNT
+            ==========================================
+            0xffff9521b3f43b80 INTERRUPTIBLE         1
+                              __schedule+0x24e
+                              schedule+0x2c
+                              schedule_timeout+0x15d
+                              __cv_timedwait_common+0xdf
+                              __cv_timedwait_sig+0x16
+                              l2arc_feed_thread+0x66
+                              thread_generic_wrapper+0x74
+                              kthread+0x121
+                              ret_from_fork+0x35
 
-            Print stacks of threads in the RUNNING state
+        Print stacks of threads in the RUNNING state
 
-                sdb> stacks -t RUNNING
-                TASK_STRUCT        STATE             COUNT
-                ==========================================
-                0xffff95214ff31dc0 RUNNING               1
+            sdb> stacks -t RUNNING
+            TASK_STRUCT        STATE             COUNT
+            ==========================================
+            0xffff95214ff31dc0 RUNNING               1
+
+        Count the number of stacks in the zfs module
+
+           sdb> stacks -m zfs | count
+           (unsigned long long)12
+
+        Print stacks of the threads started by the zthr command
+
+            sdb> threads | filter obj.comm == "zthr_procedure" | stack
+            TASK_STRUCT        STATE             COUNT
+            ==========================================
+            0xffff9c7e6c268000 INTERRUPTIBLE         5
+                              __schedule+0x24e
+                              schedule+0x2c
+                              cv_wait_common+0x118
+                              __cv_wait_sig+0x15
+                              zthr_procedure+0x45
+                              thread_generic_wrapper+0x74
+                              kthread+0x121
+                              ret_from_fork+0x1f
+
+            0xffff9c7e6c1f8000 INTERRUPTIBLE         1
+                              __schedule+0x24e
+                              schedule+0x2c
+                              schedule_hrtimeout_range_clock+0xb9
+                              schedule_hrtimeout_range+0x13
+                              __cv_timedwait_hires+0x117
+                              cv_timedwait_hires_common+0x4b
+                              cv_timedwait_sig_hires+0x14
+                              zthr_procedure+0x96
+                              thread_generic_wrapper+0x74
+                              kthread+0x121
+                              ret_from_fork+0x1f
+
     """
 
-    names = ["stacks"]
+    names = ["stacks", "stack"]
+    input_type = "struct task_struct *"
+    output_type = "struct task_struct *"
+
+    def __init__(self, args: str = "", name: str = "_") -> None:
+        super().__init__(args, name)
+        self.mod_start, self.mod_end = 0, 0
+        self.func_start, self.func_end = 0, 0
+        self.match_state = ""
 
     @classmethod
     def _init_parser(cls, name: str) -> argparse.ArgumentParser:
@@ -190,6 +212,40 @@ class Stacks(sdb.Command):
         exit_state = task.exit_state.value_()
         return Stacks.TASK_STATES[(state | exit_state) & 0x7f]
 
+    @staticmethod
+    def resolve_state(tstate: str) -> str:
+        tstate = tstate.upper()
+        if tstate in Stacks.TASK_STATE_SHORTCUTS:
+            return Stacks.TASK_STATES[Stacks.TASK_STATE_SHORTCUTS[tstate]]
+        return tstate
+
+    @staticmethod
+    def get_frame_pcs(task: drgn.Object) -> List[int]:
+        frame_pcs = []
+        try:
+            for frame in sdb.get_prog().stack_trace(task):
+                frame_pcs.append(frame.pc)
+        except ValueError:
+            #
+            # Unwinding the stack of a running/runnable task will
+            # result in an exception. Since we expect some tasks to
+            # be running, we silently ignore this case, and move on.
+            #
+            # Unfortunately, the exception thrown in this case is a
+            # generic "ValueError" exception, so we may wind up
+            # masking other "ValueError" exceptions that are not due
+            # to unwinding the stack of a running task.
+            #
+            # We can't check the state of the task here, and verify
+            # it's in the "R" state, since that state can change in
+            # between the point where the "ValueError" exception was
+            # originally raised, and here where we'd verify the
+            # state of the task; i.e. it could have concurrently
+            # transitioned from running to some other state.
+            #
+            pass
+        return frame_pcs
+
     #
     # Unfortunately the drgn Symbol API does not specify the namelist
     # that a symbol came from. As a result, we created the following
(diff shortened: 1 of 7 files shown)


sdb/commands/stacks.py
```diff
@@ -228,74 +286,93 @@ def validate_args(self, args: argparse.Namespace) -> None:
...
+                    func_match = True
+            except LookupError:
+                func, offset = hex(frame_pc), 0x0
+
+            #
+            # As a potential future item, we may want to print
+            # the frame with the module where the pc/function
+            # belongs to. For example:
+            #     txg_sync_thread+0x15e [zfs]
+            #
```
sdimitro: You can delete this too as I issued https://github.com/delphix/sdb/issues/170

sdb/commands/stacks.py
```diff
@@ -321,56 +374,43 @@ def _call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
...
+        #
+        # This implementation only works for linux kernel targets
+        # (crash dumps or live systems). When support for userland is added we can
+        # refactor the kernel code into its own function and switch to the correct
+        # codepath depending on the target.
+        #
+        if not sdb.get_target_flags() & drgn.ProgramFlags.IS_LINUX_KERNEL:
+            raise sdb.CommandError(self.name,
+                                   "userland targets are not supported yet")
+        self.validate_args(self.args)
```
sdimitro: This part doesn't have to do anything with "gathering threads", can we wrap it in its own method (`validate_context` maybe?) and call it from both `pretty_print` and `no_input` ?

sdb/commands/stacks.py
```diff
@@ -321,56 +374,43 @@ def _call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
...
-
-                try:
-                    sym = sdb.get_symbol(frame_pc)
-                    func, offset = sym.name, frame_pc - sym.address
-                    if self.args.function and self.args.function == func:
-                        func_match = True
-                except LookupError:
-                    func, offset = hex(frame_pc), 0x0
+        if to_print:
+            self.print_header()
```
sdimitro: Can we move this only to the `pretty_printer` ?

sdb/commands/stacks.py
```diff
@@ -85,47 +65,84 @@ class Stacks(sdb.Command):
...
+                              kthread+0x121
+                              ret_from_fork+0x1f
+
     """
 
-    names = ["stacks"]
+    names = ["stacks", "stack"]
+    input_type = "struct task_struct *"
+    output_type = "struct task_struct *"
+    mod_start, mod_end = -1, -1
```
sdimitro: what would happen if you removed this line and hit the test cases of filtering by module that exists and the bogus module?

shartse: Do you mean removing the -1 initial values from the variable or removing the variables all together?

shartse: I see. Removing the declaration entirely and just creating the variables if the module is defined. I tested that and it works for both valid and bogus modules. Does that seem cleaner to you?

shartse: Note: the lint check doesn't like it you create instance variables on demand. I added an `init` function, initializing the variables there.

tests/integration/data/regression_output/linux/stacks -t bogus
```diff
@@ -1 +1 @@
-sdb: stacks: 'bogus' is not a valid task state (acceptable states: RUNNING, INTERRUPTIBLE, UNINTERRUPTIBLE, STOPPED, TRACED, DEAD, ZOMBIE, PARKED, IDLE)
+sdb: stacks: 'BOGUS' is not a valid task state (acceptable states: RUNNING, INTERRUPTIBLE, UNINTERRUPTIBLE, STOPPED, TRACED, DEAD, ZOMBIE, PARKED, IDLE)
```
sdimitro: this diff is because the argument passed by the  user was mutated (see other comment)

sdb/commands/stacks.py
```diff
@@ -216,86 +248,120 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
-        # Resolve TSTATE shortcut and/or sanitize it to standard uppercase
-        # notation if it exists.
-        #
         if self.args.tstate:
-            if self.args.tstate in Stacks.TASK_STATE_SHORTCUTS:
-                self.args.tstate = Stacks.TASK_STATES[
-                    Stacks.TASK_STATE_SHORTCUTS[self.args.tstate]]
-            else:
-                self.args.tstate = self.args.tstate.upper()
+            self.args.tstate = Stacks.resolve_state(self.args.tstate)
```
sdimitro: In general it is better to use local variables and not mutate `self.args` variables themselves as this can introduce unexpected behavior in error messages and other places where they are used. Can we use a local variable for tstate here? (the error output should still use `self.args.tstate`)

sdb/commands/stacks.py
```diff
@@ -190,6 +211,17 @@ def task_struct_get_state(task: drgn.Object) -> str:
         exit_state = task.exit_state.value_()
         return Stacks.TASK_STATES[(state | exit_state) & 0x7f]
 
+    #
+    # Resolve TSTATE shortcut and/or sanitize it to standard uppercase
+    # notation if it exists.
+    #
+    @staticmethod
+    def resolve_state(tstate: str) -> str:
+        tstate = tstate.upper()
```
sdimitro: Besides the other place mentioned below where `self.args.tstate` is set directly to whatever is returned by this function, it also seems like this function would mutate its arguments anyway with this statement.

sdb/commands/stacks.py
```diff
@@ -85,47 +65,88 @@ class Stacks(sdb.Command):
...
 
-    names = ["stacks"]
+    names = ["stacks", "stack"]
+    input_type = "struct task_struct *"
+    output_type = "struct task_struct *"
+
+    def __init__(self, args: str = "", name: str = "_") -> None:
+        super().__init__(args, name)
+        self.mod_start = 0
+        self.mod_end = 0
```
sdimitro: (nit): since these two variables are related you could do the shorthand `self.mod_start, self.mod_end = 0, 0` - I don't feel strongly about this btw so feel free to drop it

sdb/commands/stacks.py
```diff
@@ -216,86 +248,120 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
+            except LookupError:
+                func, offset = hex(frame_pc), 0x0
+
+            stacktrace_info += "{:18s}{}+{}\n".format("", func, hex(offset))
+        if (self.args.function and not func_match) or (self.args.module and
+                                                       not mod_match):
+            match = False
+        if match:
+            print_cb(stacktrace_info)
+        return match
```
sdimitro: It seems like `match` starts as `True` and gets set to `False` only in the above if-clause. Why don't we get rid of it and do something like this:
```
if (.....):
    return False
print_cb(stacktrace_info)
return True
```

sdb/commands/stacks.py
```diff
@@ -321,56 +387,33 @@ def _call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
...
+        #
+        # gather_threads is a generator, but since pretty_print is not, we
+        # convert the output to a list to drain it.
+        #
+        list(self.gather_threads(objs, print))
+
+    def no_input(self) -> Iterable[drgn.Object]:
+        self.validate_context()
+        yield from self.gather_threads(for_each_task(sdb.get_prog()),
+                                       lambda x: None)
```
sdimitro: Wouldn't it be better if we made the declaration of `gather_threads()` have `print_cb` to be `None` as the default/optional argument, this way you wouldn't have to pass the lambda here.

This may give errors but I was thinking something like this:
```
    def gather_threads(
            self, objs: Iterable[drgn.Object],
            print_cb: Optional[Callable[[str], None]] = None) -> Iterable[drgn.Object]:
....
    if print_cb:
        print_cb(stacktrace_info)
    return True
```

Then `no_input` should work like this:
```
yield from self.gather_threads(for_each_task(sdb.get_prog()))
```

I think it is fine leaving it as is too, I just though that the lambda was a bit awkward.

sdb/commands/stacks.py
```diff
@@ -216,86 +248,120 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
+        # the assignment error below.
         #
+        frame_pcs: Tuple[int] = stack_key[1:]  # type: ignore[assignment]
+        for frame_pc in frame_pcs:
+            if self.args.module and self.mod_start <= frame_pc < self.mod_end:
+                mod_match = True
+
+            try:
+                sym = sdb.get_symbol(frame_pc)
+                func, offset = sym.name, frame_pc - sym.address
```
ahrens: [nit] use 2 lines of code to initialize these two variables

sdb/commands/stacks.py
```diff
@@ -216,86 +248,120 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
+        # but this is not easily expressed in mypy, thus we ignore
+        # the assignment error below.
         #
+        frame_pcs: Tuple[int] = stack_key[1:]  # type: ignore[assignment]
+        for frame_pc in frame_pcs:
+            if self.args.module and self.mod_start <= frame_pc < self.mod_end:
+                mod_match = True
+
+            try:
+                sym = sdb.get_symbol(frame_pc)
```
ahrens: From our previous discussion, it sounded like `sdb.get_symbol()` was slow (at least when called on every frame of every stack).  If we don't care about printing, and aren't filtering on function, we could avoid this call.  Even if we are filtering on function, we could avoid this call by saving the function's offset and length, like we do for the module.

Once we've avoided symbol lookup when we aren't printing, it would be performant to separate filtering from printing, which I think would be a nice refactoring of this code.  It would probably also be performant to do filtering before aggregation, which would more cleanly separate finding/filtering from aggregating/printing.

For an example of filtering and printing separated, see dbuf.py's no_input, pretty_print, and argfilter methods.

sdb/commands/stacks.py
shartse: So I can add a local variable, but the original implementation of stacks does mutate self.args.state. Are you suggesting that we re-work that entirely? 

I don't think it's inherently bad that the error message gives an upper case version of what was passed (all the options it suggestions for valid states are upper case).

sdb/commands/stacks.py
ahrens: Despite our conversation the other day, I don't really understand the benefit of passing the callback, vs. passing a boolean, especially if gather_threads() has to check the argument in either case.

All that said, it would be really nice if we could redesign this such that we can separate gathering from printing (see my other comments on how this could be made to perform well).

sdb/commands/stacks.py
ahrens: I think that in python, strings are constant, so `tstate.upper()` doesn't modify `tstate`, and `tstate = ...` doesn't modify the caller's variable, just reassigns the local variable to a different value.

sdb/commands/stacks.py
shartse: I like that idea, thanks for the suggestion!

sdb/commands/stacks.py
```diff
@@ -216,86 +248,120 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
             header += " {:>6s}".format("COUNT")
         print(header)
         print("=" * 42)
 
+    #
+    # Return a boolean indicating if this stack matches the state, function or module
+    # arguments (if any) and a string representation of the stack.
+    #
+    def match_stack(self, stack_key: Tuple[Any], tasks: List[drgn.Object],
+                    print_cb: Callable[[str], None]):
```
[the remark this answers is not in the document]
shartse: It's actually no longer returned. I'll correct the comment

sdb/commands/stacks.py
```diff
@@ -216,86 +248,120 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
+        match, stacktrace_info = True, ""
```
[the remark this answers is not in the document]
shartse: I removed `match`, so we're good.

sdb/commands/stacks.py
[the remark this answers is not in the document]
shartse: Oddly, I haven't been able to figure out how to use `Bool` (or `bool` or `Boolean`) as a type annotation. Importing any of them from the `typing` module has failed. I can look some more in the docs to figure out if there's a way to do this.

sdb/commands/stacks.py
```diff
@@ -216,86 +248,120 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
-    def _call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
-        # pylint: disable=too-many-locals
-        # pylint: disable=too-many-branches
-        # pylint: disable=too-many-statements
+    @staticmethod
+    def find_mod_offset(module: str) -> Tuple[int, int]:
+        mod_start, mod_size = Stacks.find_module_memory_segment(module)
+        assert mod_start != -1
+        mod_end = mod_start + mod_size
+        return mod_end, mod_start
```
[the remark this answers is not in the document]
shartse: Getting rid of it sounds good to me.

sdb/commands/stacks.py
```diff
@@ -321,56 +387,33 @@ def _call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
...
-            print(stacktrace_info)
-        return []
+    def pretty_print(self, objs: Iterable[drgn.Object]) -> None:
+        self.validate_context()
+        self.print_header()
+        #
+        # gather_threads is a generator, but since pretty_print is not, we
+        # convert the output to a list to drain it.
+        #
+        list(self.gather_threads(objs, print))
```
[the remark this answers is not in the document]
shartse: I like `deque` - I think the verb name makes it clearer what we're doing. Thanks for the suggestion!

sdb/commands/stacks.py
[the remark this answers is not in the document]
shartse: Figured it out - looks like I can just use the built in.

sdb/commands/stacks.py
shartse: Discussed offline: I'm adding a new instance variable so we don't mutate what was passed in args.

sdb/commands/stacks.py
shartse: I updated the PR to incorporate these suggestions (with the caveat that we still need to filter in the printing case).

Let me know what you think!

sdb/commands/stacks.py
```diff
@@ -216,120 +272,102 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
             raise sdb.CommandError(self.name,
                                    "userland targets are not supported yet")
-        self.validate_args(self.args)
+        self.validate_args()
+
+    def validate_args(self) -> None:
+        if self.args.function:
+            try:
+                func = sdb.get_object(self.args.function)
+                sym = sdb.get_symbol(func.address_of_())
```
ahrens: I saw https://github.com/osandov/drgn/issues/47 - does that mean that `-c func` will only work with non-`static` symbols now?

sdb/commands/stacks.py
```diff
@@ -338,39 +376,26 @@ def _call(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
...
-            if mod_start != -1 and not mod_match:
-                continue
-            if self.args.function and not func_match:
-                continue
             print(stacktrace_info)
-        return []
+
+    def filter_threads(self,
+                       objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
+        return (task for task in objs if self.match_stack(task))
```
ahrens: I think you can use the built-in `filter()`. i.e. `filter(objs, self.match_stack)` instead of `filter_threads(objs)`.  This also has the advantage of not creating the list (it's a generator IIRC).

sdb/commands/stacks.py
shartse: So I was originally trying to look up the symbol with the function name: `prog.symbol(self.args.function)`, which is when I hit the issue with `static` functions. Serapheim came up with this strategy as a workaround. Since switching to using the address, all functions I've tried have worked fine.

Serapheim suggested linking osandov/drgn#47 in a comment in the code to explain why we do this somewhat roundabout lookup, I think I'll do that.

shartse: tldr: `-c func` should work the same as ever.

sdb/commands/stacks.py
shartse: Good idea, that's cleaner. Though I don't think I'm creating a list here - the `(...)` syntax is creating a generator as well.

sdb/commands/stacks.py
sdimitro: I think a comment between the `try:` and the `func =` line would be nice explaining that due to https://github.com/osandov/drgn/issues/47 we can't do `sym = sdb.get_symbol(self.args.function)`  so we end up first resolving the address from the object and then getting the symbol from that address (instead of the symbol name).

shartse: I also added a new test for `-c zthr_procedure` to make sure that static functions continue to work.

sdb/commands/stacks.py
```diff
@@ -216,120 +272,106 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
+            if self.args.module and self.mod_start <= frame_pc < self.mod_end:
+                mod_match = True
+
+            if self.args.function and self.func_start <= frame_pc < self.func_end:
+                func_match = True
+
+        if (self.args.function and not func_match) or (self.args.module and
+                                                       not mod_match):
+            return False
+        return True
```
sdimitro: Not sure if that'd be too much but we could short-circuit the flow here a bit. I thought of 2 things:

[1] Iterate through the frames only if we're actually asked to filter by module or function.
[2] If everything has matched up to now, don't bother looking at subsequent frames.

```
    def match_stack(self, task: drgn.Object) -> bool:
        if self.args.tstate and self.match_state != Stacks.task_struct_get_state(task):
            return False

        if not (self.args.module or self.args.function): # point [1]
            return True

        mod_match, func_match = not self.args.module, not self.args.function # point [2]
        for frame_pc in Stacks.get_frame_pcs(task):
            if not mod_match and self.mod_start <= frame_pc < self.mod_end:
                mod_match = True

            if not func_match and self.func_start <= frame_pc < self.func_end:
                func_match = True

            if mod_match and func_match:
                return True
        return False
```

Probably splitting hairs here so feel free to drop any/both of the above.

sdb/commands/stacks.py
sdimitro: Good idea!

sdb/commands/stacks.py
ahrens: oh cool, I didn't realize thats what it did

sdb/commands/stacks.py
```diff
@@ -216,120 +272,104 @@ def find_module_memory_segment(mod_name: str) -> Tuple[int, int]:
...
-        #
+    #
+    # De-duplicate the objs (task_structs) using a dictionary indexed by
+    # task state and program counters. Return a collection sorted by number
+    # of tasks per stack.
+    #
+    @staticmethod
+    def aggregate_stacks(
+            objs: Iterable[drgn.Object]) -> List[Tuple[Any, List[drgn.Object]]]:
         stack_aggr: Dict[Any, List[drgn.Object]] = defaultdict(list)
```
ahrens: [nit] since stack_key is now a fixed-length tuple, each entry of known type (yay!), can we be more specific than `Any` on these 2 lines?  Would `Tuple[str, Tuple[int]]` work?  (not sure how the variable-length Tuple-of-ints should be specified, it may have to fall back on Any)

shartse: I looked this up, looks like we can do something similar:

>  Tuples can also be used as immutable, varying-length sequences. You can use the type Tuple[T, ...] (with a literal ... – it’s part of the syntax) for this purpose

So I can do `Tuple[str, Tuple[int, ...]]`

PR #165, opened 2020-02-18
Locators should only use no_input functionality when they are first in a pipeline

### LINKED ISSUE #151: filter sometimes passes all objects when condition is not met
**Expected Behavior**
When using a filter with a `Locator`, I would expect that only the objects matching the filter are returned, for example:
```
sdb> slabs | filter obj.name == "kmalloc-512" | slabs | count
(unsigned long long)1
```
If no objects match, I would expect no output. 

**Actual Behavior**
Instead, all objects pass through the filter:
```
sdb> slabs | filter obj.name == "bogus" | slabs | count
(unsigned long long)132
```
Interestingly, this does not reproduce with a simple base case of filter:
```
sdb> echo 0x0  0x2 | filter obj == 1
>
sdb> echo 0x0  0x2 0x1 | filter obj == 1
(void *)0x1
```

shartse:
## Pull request checklist

- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Build was run locally and any changes were pushed
- [X] Lint has passed locally and any fixes were made for failures

## Pull request type
resolves #151 
 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

`Locators` invoke their `no_input` functionality when the command receives no input. This can lead to some unexpected behavior where empty filters appear to match all objects:
 
```
sdb> slabs | filter obj.name == "bogus" | slabs | count
(unsigned long long)131
```
## What is the new behavior?
Now `Locators` will not produce new input mid-pipeline when they are pre-fixed by a filter with no output. 
```
sdb> slabs | filter obj.name == "bogus" | slabs | count
(unsigned long long)0
```

I couldn't find any documented use cases or tests that relied on the previous nehavior

- [ ] Yes
- [X] No

### DIFF
--- a/sdb/command.py
+++ b/sdb/command.py
@@ -217,6 +217,7 @@ class Command:
 
     def __init__(self, args: str = "", name: str = "_") -> None:
         self.name = name
+        self.isfirst = False
         self.islast = False
 
         self.parser = type(self)._init_parser(name)
@@ -681,9 +682,12 @@ class Locator(Command):
             baked[type_canonicalize_name(
                 method.input_typename_handled)] = method
 
-        has_input = False
+        if self.isfirst:
+            assert not objs
+            yield from self.no_input()
+            return
+
         for i in objs:
-            has_input = True
             obj_type_name = type_canonical_name(i.type_)
 
             # try subclass-specified input types first, so that they can
@@ -711,8 +715,6 @@ class Locator(Command):
             # error
             raise CommandError(
                 self.name, 'no handler for input of type {}'.format(i.type_))
-        if not has_input:
-            yield from self.no_input()
 
     def _call(self,
               objs: Iterable[drgn.Object]) -> Optional[Iterable[drgn.Object]]:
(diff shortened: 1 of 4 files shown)


sdb/command.py
```diff
@@ -711,7 +712,8 @@ def caller(self, objs: Iterable[drgn.Object]) -> Iterable[drgn.Object]:
             # error
             raise CommandError(
                 self.name, 'no handler for input of type {}'.format(i.type_))
-        if not has_input:
+        if self.isfirst:
+            assert not has_input
```
ahrens: I think it would be more clear to structure this code like so, in which case we can remove the `has_input` variable:

```
if self.isfirst:
  assert that objs is empty
  yield from...
else:
  .. set up out_type ..
  .. set up baked ..
  for i in objs:
    ...
```

or
```
if self.isfirst:
  assert that objs is empty
  yield from...
  return
... set up out_type...
etc
```

tests/integration/test_core_generic.py
```diff
@@ -70,6 +70,8 @@
     "spa rpool | filter obj.spa_syncing_txg <= 1624 | member spa_name",
     "spa rpool | filter obj.spa_syncing_txg < 1624 | member spa_name",
     "spa rpool | filter obj.spa_syncing_txg > 1624 | member spa_name",
+    # locator that recieves no input from a filter
```
sdimitro: (nit) s/recieves/receives

thanks for writing thorough test cases!

PR #171, opened 2020-02-20
Introduce slub cache walker

sdimitro:
Note: one-line change as a side commit - gitignore *.swp vim files


sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +124,173 @@ def util(cache: drgn.Object) -> int:
...
+    """
+    Return a set of addresses that point to all the freed
+    objects in this slab.
+    """
+    free_objects = []
+    p = page.freelist
+    while p != sdb.get_typed_null(p.type_):
+        free_objects.append(p.value_())
+        p = cache_get_free_pointer(cache, p)
+    return set(free_objects)
```
ahrens: not sure if this really matters, but I was wondering if it's "pythonic" to construct the set from an anonymous generator.  Not sure of the syntax, but I'm imagining something like:

```
return set({
  p = page.freelist
  while ... :
    yield p.value_()
   p = cache_get_free_pointer...
})
```

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +124,173 @@ def util(cache: drgn.Object) -> int:
...
+
+
+def cache_gather_children(cache: drgn.Object) -> List[drgn.Object]:
+    children = []
+    for child in list_for_each_entry("struct kmem_cache",
+                                     cache.memcg_params.children.address_of_(),
+                                     "memcg_params.children_node"):
+        children.append(child)
+    return children
```
ahrens: same thing here, could this return an `Iterable` rather than a `List`, and be a generator?

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +124,173 @@ def util(cache: drgn.Object) -> int:
...
+        return set(partials)
+
+    for child in list_for_each_entry("struct kmem_cache",
+                                     cache.memcg_params.children.address_of_(),
+                                     "memcg_params.children_node"):
+        partials += cache_get_partials(child)
+    return set(partials)
+
+
+def for_each_object_in_cache(cache: drgn.Object) -> drgn.Object:
```
ahrens: I think this returns `Iterable[drgn.Object]`

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +124,173 @@ def util(cache: drgn.Object) -> int:
...
+            # where subsequently filtered out. Another one is
+            # that these regions are just not mapped to anything.
+            # In any case this scenario hasn't brought up any
+            # issues with the functionality that we are trying
+            # to provide.
+            #
+            continue
+
+        slab_cache = page.slab_cache
+        if slab_cache == sdb.get_typed_null(slab_cache.type_):
```
ahrens: Is there a more elegant way to do this, something like `slab_cache.value_() == 0` ?  The docs say:
> For pointers, this [.value_()] returns the address value of the pointer.

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +124,173 @@ def util(cache: drgn.Object) -> int:
...
+                free_set = cache_get_free_set(cache, page)
+            except drgn.FaultError as err:
+                #
+                # It is common for freelists of partials slabs to be
+                # updated while we are looking at them. Instead of
+                # stopping right away or going though and reading
+                # more invalid data we just skip to the next slab.
+                #
+                print(
+                    f"inconsistent freelist dereference {hex(err.address)} - page {hex(page)}"
```
ahrens: I'd suggest a printout format that can't be confused with subtraction.  e.g. `address: {hex(err.address)} page: {hex(page)}`

sdb/commands/linux/slabs.py
```diff
@@ -197,3 +202,44 @@ def pretty_print(self, objs: Iterable[drgn.Object]) -> None:
...
+
+
+class SlubCacheWalker(sdb.Walker):
+    """
+    Walk though all allocated entries in a slub cache.
+
+    EXAMPLES
+        Walk through all the objects in the TCP cache:
+
+            sdb> slabs | filter obj.name == "TCP" | slub_cache
```
ahrens: I think that since this is a Walker, one of the examples could use ` | walk`

sdb/commands/linux/internal/slub_helpers.py
sdimitro: Per our conversation I created a local generator function instead of the list. Then pylint brought up that the actual compression wasn't needed to create the set and using the `set()` constructor is preferred, thus I ended up with this:
```
def cache_get_free_set(cache: drgn.Object, page: drgn.Object) -> Set[int]:
    """
    Return a set of addresses that point to all the freed
    objects in this slab.
    """

    def __cache_get_free_set_gen() -> Iterable[int]:
        p = page.freelist
        while p.value_() != 0x0:
            yield p.value_()
            p = cache_get_free_pointer(cache, p)

    return set(__cache_get_free_set_gen())
```

sdb/commands/linux/internal/slub_helpers.py
sdimitro: Changed to a Set[int] as discussed.

sdb/commands/linux/slabs.py
sdimitro: Good idea! Added a test-case too

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +124,174 @@ def util(cache: drgn.Object) -> int:
...
+    node = cache.node[0].partial  # assumption nr_node_ids == 0
+    for page in list_for_each_entry("struct page", node.address_of_(), "lru"):
+        partials.append(page.value_())
+
+    if not is_root_cache(cache):
+        return set(partials)
+
+    for child in list_for_each_entry("struct kmem_cache",
+                                     cache.memcg_params.children.address_of_(),
+                                     "memcg_params.children_node"):
```
ahrens: It would be nice if this could use reuse the list_for_each_entry call that's already in cache_gather_children().  Maybe if we changed that to be a generator that yields the drgn.Object's, and then the other caller could create the set from that?

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +124,174 @@ def util(cache: drgn.Object) -> int:
...
+    ptr_as_ulong = drgn.Object(sdb.get_prog(),
+                               type='unsigned long',
+                               value=ptr.value_())
+    clean_ptr_val = ptr_as_ulong.value_()
+    clean_ptr_val ^= cache.random.value_()
+    clean_ptr_val ^= ptr_addr.value_()
+    return drgn.Object(sdb.get_prog(), type='void *', value=clean_ptr_val)
+
+
+def cache_get_free_set(cache: drgn.Object, page: drgn.Object) -> Set[int]:
```
ahrens: Given that these are the helpers, which presumably we want to be fairly generic, I wonder if these should generally be generators which yield the drgn.Object's, (i.e. return `Iterable[drgn.Object]`) and then callers that want sets of ints can do `set(o.value_() for o in cache_get_freelist(...))`.  I think this applies to all of the functions that return `Set[int]` in this file.

sdimitro: I applied this feedback and also refactored some of the old code in this file that was like this. Thanks for bringing this up. It does look nicer!

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +113,160 @@ def util(cache: drgn.Object) -> int:
...
+                               type='unsigned long',
+                               value=ptr.value_())
+    clean_ptr_val = ptr_as_ulong.value_()
+    clean_ptr_val ^= cache.random.value_()
+    clean_ptr_val ^= ptr_addr.value_()
+    return drgn.Object(sdb.get_prog(), type='void *', value=clean_ptr_val)
+
+
+def cache_get_free_set(cache: drgn.Object,
+                       page: drgn.Object) -> Iterable[drgn.Object]:
```
ahrens: It seems like the convention for how to name a function that gets all X given a Y is `for_each_X_in_Y`, so I think this should be `for_each_freeobj_in_slab`.

sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -122,3 +113,160 @@ def util(cache: drgn.Object) -> int:
...
+    """
+    Return all the free objects in this slab.
+    """
+    p = page.freelist
+    while p.value_() != 0x0:
+        yield p
+        p = cache_get_free_pointer(cache, p)
+
+
+def cache_gather_partial_slabs(cache: drgn.Object) -> Iterable[drgn.Object]:
```
ahrens: `for_each_partial_slab_in_cache`?

sdb/commands/linux/internal/slub_helpers.py
sdimitro: done and good idea - we can start standardizing this now. I'll probably do this and all the yielding refactoring in spl_kmem_cache walker after this lands.

sdimitro:
> FWIW - if we are inventing the convention now, personally I prefer the brevity of `for_X_in_Y`, but I think the Linux kernel may already be using `for_each_X_in_Y`.

I can refactor all of them as a separate commit when I do the spl_kmem_caches walker.

PR #186, opened 2020-02-25
Introduce spl cache command

sdimitro:
A few side-changes added as separate commits.


sdb/commands/internal/p2.py
```diff
@@ -0,0 +1,29 @@
...
+import sdb
+
+
+def p2roundup_typed(val: int, align: int, type_: str) -> int:
+    """
+    Round up `val` to the next `align` boundary (value sized as `type_`).
+    """
+    typed_val = sdb.create_object(type_, val)
+    typed_align = sdb.create_object(type_, align)
+    return (((typed_val - 1) | (typed_align - 1)) + 1).value_()
```
ahrens: Are there any use cases where it's important that we do this with the specified type, vs. using python's default numbers?  If so, that seems pretty subtle and perhaps worthy of a comment.

sdb/commands/spl/internal/kmem_helpers.py
```diff
@@ -116,20 +121,88 @@ def entry_size(cache: drgn.Object) -> int:
...
+        if spl_util.is_list_empty(sko.sko_list):
+            yield obj
+
+
+def for_each_object_in_spl_cache(cache: drgn.Object) -> Iterable[drgn.Object]:
+    assert sdb.type_canonical_name(cache.type_) == 'struct spl_kmem_cache *'
+
+    slab_object_iterator = for_each_onslab_object_in_slab
+    if 'KMC_OFFSLAB' in list(for_each_slab_flag_in_cache(cache)):
+        slab_object_iterator = for_each_offslab_object_in_slab
```
ahrens: Why bother with  `for_each_offslab_object_in_slab`, instead of having a check here that KMC_OFFSLAB is not set?  That way you wouldn't need `slab_object_iterator`.

sdb/target.py
```diff
@@ -128,6 +128,18 @@ def type_canonicalize_name(type_name: str) -> str:
...
+def type_canonicalize_size(t: Union[drgn.Type, str]) -> int:
+    """
+    Return the "canonical size" of this type. See type_canonicalize().
+    """
+    if isinstance(t, str):
+        type_ = get_type(t)
+    else:
+        assert isinstance(t, drgn.Type)
+        type_ = t
+    return type_canonicalize(type_).size
```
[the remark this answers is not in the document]
sdimitro: Yes unfortunately

sdb/commands/internal/p2.py
sdimitro: For some reason I thought it would but trying out a few edge-cases in the `drgn`-repl I don't think that it matters. I'll use the default python.

sdb/commands/spl/internal/spl_util.py
```diff
@@ -0,0 +1,29 @@
...
+
+import drgn
+import sdb
+
+
+def is_list_empty(l: drgn.Object) -> bool:
+    """
+    True if list is empty, False otherwise.
+    """
+    assert sdb.type_canonical_name(l.type_) == 'struct list_head'
```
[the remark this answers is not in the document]
sdimitro: Moved to linux directory within the linked_list file.

sdimitro:
@ahrens this is good for another round whenever you get the chance.

sdb/commands/spl/internal/kmem_helpers.py
```diff
@@ -116,20 +121,79 @@ def entry_size(cache: drgn.Object) -> int:
...
+    for i in range(slab.sks_objs.value_()):
+        obj = sdb.create_object('void *',
+                                slab.value_() + sks_size + (i * spl_obj_size))
+        #
+        # If the sko_list of the object is empty, it means that
+        # this object is not part of the slab's internal free list
+        # and therefore it is allocated.
+        #
+        sko = sko_from_obj(cache, obj)
+        assert sko.sko_magic.value_() == 0x20202020  # SKO_MAGIC
```
ahrens: If this `assert` fails, do we get a "you hit a bug" message?  Seems like this is not necessarily a bug in sdb, so we would want it to be a CommandError.

sdb/commands/spl/internal/kmem_helpers.py
```diff
@@ -116,20 +121,79 @@ def entry_size(cache: drgn.Object) -> int:
...
+        if linked_lists.is_list_empty(sko.sko_list):
```
ahrens: My understanding is that `sko_list` is not a list, but rather a link on a list.  And `is_list_empty` is not telling us if this "list" is empty, but rather if the link is on any list.  Most of this is the fault of the linux kernel's choice of terminology, so I'm not sure how much we can do about this besides adding a comment (which is sort-of covered above but I think we could be more explicit).

sdb/commands/spl/internal/kmem_helpers.py
sdimitro: Good idea!

sdb/commands/spl/internal/kmem_helpers.py
sdimitro: Added a more explicit note about this in the existing comment.

PR #191, opened 2020-02-28
Fix latest mypy failures and enable --strict

sdimitro:
mypy: fix recent failures introduced by the new drgn types
mypy: enable mypy --strict

### DIFF
--- a/sdb/commands/linux/internal/slub_helpers.py
+++ b/sdb/commands/linux/internal/slub_helpers.py
@@ -27,7 +27,7 @@ import sdb
 
 def is_root_cache(cache: drgn.Object) -> bool:
     assert cache.type_.type_name() == 'struct kmem_cache *'
-    return cache.memcg_params.root_cache.value_() == 0x0
+    return int(cache.memcg_params.root_cache.value_()) == 0x0
 
 
 def for_each_root_cache() -> Iterable[drgn.Object]:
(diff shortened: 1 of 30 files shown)


sdb/commands/linux/internal/slub_helpers.py
```diff
@@ -55,17 +55,17 @@ def nr_slabs(cache: drgn.Object) -> int:
 
 def entries_per_slab(cache: drgn.Object) -> int:
     assert cache.type_.type_name() == 'struct kmem_cache *'
-    return cache.oo.x.value_() & 0xffff
+    return int(cache.oo.x.value_()) & 0xffff
```
ahrens: Does `value_()` not return an `int`, or is it just that we can't statically determine what type it returns, so we need to "cast" it here to make the type checking happy?

sdimitro: Not necessarily -> https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.Object.value_

It could be a `float`, `bool`, or even a `List` if we are using it on an array type.

ahrens: Right, I mean in this case it returns an int, but not all calls to value_() return an int.  So `int()` has no effect at runtime but it tells the static analysis that this is now definitely an int.

PR #197, opened 2020-03-13
Use shfmt for shell scripts in .github/scripts (#118)

sdimitro:
Commit 1: Introduce the check (Should fail testing because there are formatting errors)

Testing Failure Output:
```
--- .github/scripts/download-dump-from-s3.sh.orig
+++ .github/scripts/download-dump-from-s3.sh
@@ -1,41 +1,41 @@
 #!/bin/bash -eu
 
 #
 # Assumptions for this to work:
 # [1] This script is executed from the root of the SDB repo
 # [2] The archive downloaded from S3 is lzma-compressed
 # [3] The archive's contents have the following hierarchy:
 #           dump-data
 #           ├── dump.201912060006
 #           ├── mods
 #           │   ├── avl
 #           │   │   └── zavl.ko
 #           .....
 #           │   └── zfs
 #           │       └── zfs.ko
 #           └── vmlinux-5.0.0-36-generic
 #
 
 DATA_DIR="tests/integration/data"
...

echo "initiating download of $1 from S3 ..."
 /usr/local/bin/aws s3 cp --no-sign-request s3://sdb-regression-dumps/$1 .
-[ $? -eq 0 ]  || exit 1
+[ $? -eq 0 ] || exit 1
 
 echo "decompressing dump ..."
 tar -x --lzma -f $1
 
 echo "moving contents to tests/integration/data ..."
 mv dump-data/* $DATA_DIR
-[ $? -eq 0 ]  || exit 1
+[ $? -eq 0 ] || exit 1
 
 rmdir dump-data
-[ $? -eq 0 ]  || exit 1
+[ $? -eq 0 ] || exit 1
 
 echo "Done"
 
##[error]Process completed with exit code 1.
```

Commit 2: Fixes the above formatting errors so testing can pass

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -101,3 +101,12 @@ jobs:
       - run: python3 -m pip install mypy==0.730
       - run: python3 -m mypy --strict --allow-untyped-calls --show-error-codes -p sdb
       - run: python3 -m mypy --strict --ignore-missing-imports --show-error-codes -p tests
+  #
+  # Verify that "shfmt" ran successfully against our shell scripts.
+  #
+  shfmt:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v1
+      - run: ./.github/scripts/install-shfmt.sh
+      - run: /usr/local/bin/shfmt -d .github/scripts/
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,4 +109,4 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - run: ./.github/scripts/install-shfmt.sh
-      - run: /usr/local/bin/shfmt -d .github/scripts/
+      - run: /usr/local/bin/shfmt -d .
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -101,3 +101,12 @@ jobs:
...
       - run: python3 -m mypy --strict --ignore-missing-imports --show-error-codes -p tests
+  #
+  # Verify that "shfmt" ran successfully against our shell scripts.
+  #
+  shfmt:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v1
+      - run: ./.github/scripts/install-shfmt.sh
+      - run: /usr/local/bin/shfmt -d .github/scripts/
```
pzakha: Wouldn't it better to run it against the whole repository?

sdimitro: I don't see why not. Making the change now

sdimitro: Changed it for the whole sdb directory - I think this is worthwhile as I may introduce a few scripts outside of the `.github` directory while trying to get sdb on PyPI so that it can be installed through `pip`.

PR #198, opened 2020-03-13
Want sizeof and sum commands

sdimitro:
3 commits:
[1] Refactor a function from `ptype` to be common code - the goal is to use that for `sizeof`
[2] Add `sizeof` command
[3] Add `sum` command


sdb/commands/sizeof.py
```diff
@@ -0,0 +1,64 @@
...
+            sdb> sizeof task_struct spa_t
+            (size_t)9152
+            (size_t)9176
+
+        Figuring out the size of objects passed through a
+        pipe:
+
+            sdb> addr spa_namespace_avl | addr init_task | deref | sizeof
+            (size_t)40
+            (size_t)9152
```
ahrens: fancy!  A lot of things assume/enforce that all the objects in one stage of the pipeline are of the same type.  Not sure we want to "encourage" (by way of example) folks to deviate from that.

sdimitro: Makes sense - took out the second addr command from the example

Issue #206, opened 2020-04-08
Refactor spa, vdev, and metaslab commands


sdimitro:
These commands were hacked together during the initial prototype of sdb based on crash-python and then they were shoved into drgn-based prototype from which point they were never revisited. As a result their code is far from optimal and since they are used so often we'll want to refactor them soon. Below is a list of problems with that code:

[1] man/help messages are missing - also the descriptions of each of their command line options is lacking

[2] The `arg_string` convention used/passed between these commands is very ad-hoc and MDB-like (meaning Python should allow us to do better) - here is what I'm talking about

spa.py:
```
    def __init__(self, args: str = "", name: str = "_") -> None:
        super().__init__(args, name)
        self.arg_string = ""
        if self.args.metaslab:
            self.arg_string += "-m "
        if self.args.histogram:
            self.arg_string += "-H "
        if self.args.weight:
            self.arg_string += "-w "

...

    def pretty_print(self, spas: Iterable[drgn.Object]) -> None:
            ...
            if self.args.vdevs:
                vdevs = sdb.execute_pipeline([spa], [Vdev()])
                Vdev(self.arg_string).pretty_print(vdevs, 5)   // <---- arg_string passed here
```

then in vdev.py we do the same for metaslab.py.

[3] It is annoying that in order to print just the metaslabs for now through spa (e.g. `::spa -m` in `mdb`) we are forced to provide the `-v` option on top of `-m`. E.g.

```
```

PR #207, opened 2020-04-08
Add histogram command and plug it to spa & metaslab commands

sdimitro:
Commit:
```
Want command to print ZFS Histograms
Histogram option for spa, vdev, and metaslab commands
```

This code is not optimal but I decided to open the PR so we have something for 6.0.2.0 (see https://jira.delphix.com/browse/DLPX-67353). I opened https://github.com/delphix/sdb/issues/206 for the leftover work (spa,vdev,metaslab refactoring).

Note:
I thought about making this a pretty printer but unfortunately we can't specify an generic `uint64_t` array as an input type, in a way that drgn understands (maybe we can add our own layer a weaker type equality where comparing arrays of the same type yields true regardless of whether they have the same length or not). I could create two pretty printers for these histograms though a 64 and a 32 bucket version, If people feel strongly let me know and I can change this.

Other Ideas:
Would people be interested in additional behavior where if a metaslab is loaded we print the rt_histogram of the ms_allocatable instead and we calculate the precise median from the range tree?


sdimitro:
cc @grwilson

sdb/commands/zfs/histograms.py
```diff
@@ -0,0 +1,149 @@
...
+            if space_left <= space_in_bucket:
+                median = 1 << (bucket + offset)
+                #
+                # Size of segments may vary within one bucket thus we
+                # attempt to approximate the median by looking at the
+                # number of segments in the bucket and assuming that
+                # they are evenly distributed along the bucket's range.
+                #
+                bucket_fill_pct = (space_left * 100) / space_in_bucket
+                median += round(median * bucket_fill_pct / 100)
```
ahrens, line 103: In python3, integer division yields a floating-point result.  So I don't think we gain anything by doing this as a "percent" (`*100`).  You could do `round(median * space_left / space_in_bucket)` or
```
bucket_fill = space_left / space_in_bucket
median += round(median * bucket_fill)
```

sdb/commands/zfs/histograms.py
```diff
@@ -0,0 +1,149 @@
...
+        Returns the approximated median of a ZFS histogram.
+        """
+        canonical_type = sdb.type_canonicalize(hist.type_)
+        assert canonical_type.kind == drgn.TypeKind.ARRAY
+        assert sdb.type_canonicalize(
+            canonical_type.type).kind == drgn.TypeKind.INT
+        total_space, max_bucket, min_bucket = 0, 0, (len(hist) - 1)
+
+        for (bucket, value) in enumerate(hist):
+            count = int(value.value_())
```
ahrens, line 84: I think that we don't need `.value_()` here (and elsewhere).  From the drgn docs:

> The standard int(), float(), and bool() functions convert an object to that Python type. Conversion to bool uses the programming language’s notion of “truthiness”. Additionally, certain Python functions will automatically coerce an object to the appropriate Python type (e.g., hex(), round(), and list subscripting).

sdb/commands/zfs/histograms.py
```diff
@@ -0,0 +1,149 @@
...
+                        offset: int = 0,
+                        indent: int = 0) -> None:
+        canonical_type = sdb.type_canonicalize(hist.type_)
+        assert canonical_type.kind == drgn.TypeKind.ARRAY
+        assert sdb.type_canonicalize(
+            canonical_type.type).kind == drgn.TypeKind.INT
+
+        HISTOGRAM_WIDTH_MAX = 40
+
+        max_, max_bucket, min_bucket = 0, 0, (len(hist) - 1)
```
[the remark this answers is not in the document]
sdimitro: I can't find anything online on what is considered "pythonic" for this but I personally find it more concise (if this is the right word to describe it) to group together related variables this way - assuming there are less than 4.

In this case  I had some variables in previous iterations of the code that I ended up deleting so I cal split the above (and also move `HISTOGRAM_WIDTH_MAX` closer to its usage)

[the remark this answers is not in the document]
ahrens: Yeah it's definitely more concise, but arguably less readable.

sdb/commands/zfs/histograms.py
```diff
@@ -0,0 +1,149 @@
...
+    @staticmethod
+    def print_histogram(hist: drgn.Object,
+                        offset: int = 0,
+                        indent: int = 0) -> None:
+        canonical_type = sdb.type_canonicalize(hist.type_)
+        assert canonical_type.kind == drgn.TypeKind.ARRAY
+        assert sdb.type_canonicalize(
+            canonical_type.type).kind == drgn.TypeKind.INT
+
+        max_ = 0
```
ahrens, line 122: [nit] I assume the trailing underscore is to avoid a conflict with the built-in `max()` function.  It might be cleaner to do that by renaming this to `max_value` or `max_count`

sdb/commands/zfs/metaslab.py
```diff
@@ -165,13 +165,16 @@ def pretty_print(self,
...
-            if not self.args.histogram and not self.args.weight:
+            if not self.args.weight:
                 Metaslab.print_metaslab(msp, first_time, indent)
             if self.args.histogram:
                 spacemap = msp.ms_sm
                 if spacemap != sdb.get_typed_null(spacemap.type_):
                     histogram = spacemap.sm_phys.smp_histogram
-                    print_histogram(histogram, 32, spacemap.sm_shift)
+                    ZFSHistogram.print_histogram(
+                        histogram, int(spacemap.sm_shift.value_()), indent)
```
ahrens, line 175:
```suggestion
                        histogram, int(spacemap.sm_shift), indent)
```

sdb/commands/zfs/metaslab.py
```diff
@@ -165,13 +165,16 @@ def pretty_print(self,
...
+                    ZFSHistogram.print_histogram_median(
+                        histogram, int(spacemap.sm_shift.value_()), indent)
```
ahrens, line 177:
```suggestion
                        histogram, int(spacemap.sm_shift), indent)
```

sdb/commands/zfs/histograms.py
```diff
@@ -0,0 +1,149 @@
...
+               1.0MB:    675 ***
+               2.0MB:    486 **
+               4.0MB:    267 *
+               8.0MB:    110
+              16.0MB:     50
+              32.0MB:     18
+              64.0MB:      8
+             128.0MB:     11
+             256.0MB:    102
+            Histogram Median: 339.7MB
```
prakashsurya, line 61: Since this value is an approximation, perhaps we should state that when it's displayed? e.g. `Approx. Histogram Median:` or `Approx. Median:` (since "histogram" can probably be implied at this point)..

sdb/commands/zfs/histograms.py
```diff
@@ -0,0 +1,149 @@
...
+
+    NOTE
+        The median is just an approximation as we can't tell the
+        exact size of each bucket within a histogram bucket.
+
+    EXAMPLES
+        Dump the histogram of the normal metaslab class of the rpool:
+
+            sdb> spa rpool | member spa_normal_class.mc_histogram | zhist
+            seg-size   count
```
prakashsurya, line 39: Is this command intended to be generic enough to be used with any array of integers? If not, what is the dependency on ZFS? If so, does the `seg-size` name make sense for other (e.g. non metaslab) contexts? Along the same lines, does the `zhist` name make sense, if this can be used with inputs unrelated to ZFS?

sdb/commands/zfs/histograms.py
sdimitro: Will change to `Approx. Median:`

sdb/commands/zfs/histograms.py
sdimitro: I do believe that there is more generic functionality that can come out of this but given that we are printing the indexes as nicenum'ed values shifted by an offset I'd say the command itself is ZFS specific.

We could create a generic function that optionally hides unused buckets on both ends of the histogram as we do and allows you to give it callback to manipulate the index in some way before printing it.

That said abstracting all of this seems a bit too much to me as this is a few lines of code, and most of its complexity would be in preprocessing and making sure the spacing looks nice (e.g. like `sdb/commands/internal/table.py`). Since I haven't come across any other use case where something like that would be useful I am leaning towards leaving this as is but if people think it would be useful I can give this a shot.

PR #223, opened 2020-05-12
Use "delphix/actions" for shfmt

prakashsurya:


prakashsurya:
This is depends on: https://github.com/delphix/actions/pull/3 and https://github.com/delphix/actions/pull/4

.github/workflows/main.yml
```diff
@@ -118,5 +118,4 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
-      - run: ./.github/scripts/install-shfmt.sh
-      - run: /usr/local/bin/shfmt -d .
+      - uses: delphix/actions/shfmt@master
```
pzakha, line 121: We don't also run shellcheck for sdb?

prakashsurya: I don't see it used anywhere in this repository, do you? Did I overlook something?

pzakha: Sorry, I meant that as a generic question. @sdimitro Is there an explicit reason we aren't running shellcheck? Should we add it?

sdimitro: No explicit reason - I just only knew about `shfmt` at the time when I made that change :-)

pzakha: Okay, we can perhaps consider adding it in a later commit.

PR #267, opened 2021-03-15
Add a whatis command to sdb

PaulZ-98:
## Pull request checklist

- [x] Build was run locally and any changes were pushed
- [x] Lint has passed locally and any fixes were made for failures

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There is no whatis command in sdb.

Issue Number: 119

## What is the new behavior?
sdb> whatis 0xfffffe06596e21b8
0xfffffe06596e21b8 is allocated from dmu_buf_impl_t
sdb> dbuf |head 1 |deref |member db_buf |whatis
0xffff8e804368ac80 is allocated from arc_buf_t

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/sdb/commands/linux/whatis.py
+++ b/sdb/commands/linux/whatis.py
@@ -1,0 +1,77 @@
+#
+# Copyright 2020 Delphix
+# Copyright 2021 Datto Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pylint: disable=missing-docstring
+
+import argparse
+from typing import Iterable
+
+import drgn
+import sdb
+from sdb.commands.linux.internal import slub_helpers as slub
+
+
+class WhatIs(sdb.Command):
+    """
+    Print the name of the kmem cache from which the address is allocated.
+
+    DESCRIPTION
+       The address can be specified as an argument, or
+       passed through a pipe.
+
+    EXAMPLES
+        Determine the kmem_cache for a given address:
+
+            sdb> whatis 0xfffffe06596e21b8
+            0xfffffe06596e21b8 is allocated from dmu_buf_impl_t
+
+        Determine the kmem_cache of address passed through a
+        pipe:
+
+            sdb> dbuf |head 1 |deref |member db_buf |whatis
+            0xffff8e804368ac80 is allocated from arc_buf_t
+    """
+
+    names = ["whatis"]
+
+    @staticmethod
+    def print_cache(cache: drgn.Object, addr: str) -> None:
+        if cache is None:
+            print(f"{addr} does not map to a kmem_cache")
+        else:
+            assert sdb.type_canonical_name(cache.type_) == 'struct kmem_cache *'
+            cache_nm = cache.name.string_().decode('utf-8')
+            print(f"{addr} is allocated from {cache_nm}")
+
+    @classmethod
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        parser = super()._init_parser(name)
+        parser.add_argument("address", nargs="*", metavar="<address>")
+        return parser
+
+    def _call(self, objs: Iterable[drgn.Object]) -> None:
+        for obj in objs:
+            cache = slub.lookup_cache_by_address(obj)
+            self.print_cache(cache, hex(int(obj)))
+        for addr in self.args.address:
+            try:
+                obj = sdb.create_object("void *", int(addr, 16))
+            except ValueError:
+                print(f"{addr} is not a valid address")
+                continue
+            cache = slub.lookup_cache_by_address(obj)
+            self.print_cache(cache, addr)
--- a/tests/integration/test_linux_generic.py
+++ b/tests/integration/test_linux_generic.py
@@ -94,6 +94,13 @@ POS_CMDS = [
     "threads | count",
     'threads | filter \'obj.comm == "java"\' | threads',
     "thread",
+
+    # whatis
+    "dbuf |head 1 |deref |member db_buf |whatis",
+    "whatis 0xffffa089407ca870",
+    "whatis 0xffffa0888c766000 0xffffa089407ca870",
+    "whatis 0xffff",
+    "whatis 0xf987kkbbh"
 ]
 
 STRIPPED_POS_CMDS = [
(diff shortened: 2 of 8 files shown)


PaulZ-98:
Looks like the CI is just waiting.  I had a good run, made a few changes and re-pushed a few consecutive times, and it went into the waiting state.  Can I trigger it to re-run?  I already tried a trivial commit/push  to change the commit hash.

ahrens:
@PaulZ-98 In my experience, force-pushing a different commit (e.g. with the timestamp changed) has kicked off a new test run.  So if that isn't working I'm not sure what to do.  @sdimitro can probably help when he gets back from vacation later this week.

sdb/commands/linux/whatis.py
```diff
@@ -0,0 +1,78 @@
...
+    @classmethod
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+        parser = super()._init_parser(name)
+        parser.add_argument("address", nargs="*", metavar="<address>")
+        return parser
+
+    def _call(self, objs: Iterable[drgn.Object]) -> None:
+        for obj in objs:
+            addr_str = "%s" % hex(int(obj))
+            lookup_in_kmem(addr_str)
```
ahrens, line 76: This seems a bit roundabout to me.  We're taking a drgn.Object, converting it to number (assuming its an address) and then to a hex string, and then lookup_in_kmem() is converting the string to a number and then a drgn.Object.  Maybe the helper (`lookup_in_kmem`) could take a drgn.Object, and then in the `addr` case below we could convert the string argument into a Object.

ahrens (commented):
Nice functionality, thanks for contributing this!

sdb/commands/linux/whatis.py
PaulZ-98: Thanks Matt.  Cleaning this up will give me a reason to force push a commit and hopefully get the CI running.

PaulZ-98:
drgn has modified its required version of libelf to 0.176.  It looks like Ubuntu bionic only goes up to 0.170.

PaulZ-98:
I edited install-drgn.sh to work around the libelf version problem (git checkout older drgn).  But that's not likely to be the ultimate solution.

.github/scripts/install-drgn.sh
```diff
@@ -8,7 +8,7 @@ sudo apt update
 sudo apt install bison flex libelf-dev libdw-dev libomp5 libomp-dev
 
 git clone https://github.com/osandov/drgn.git
-
 cd drgn
+git checkout -b drgn_february 0a6aaaae5d31ed142448f8220e208d65478ec80d
```
sdimitro, line 12: Yeah as you mentioned this should probably go. I think Prakash got something working that we should be able to integrate soon.

tests/integration/test_linux_generic.py
```diff
@@ -94,6 +94,13 @@
     "threads | count",
     'threads | filter \'obj.comm == "java"\' | threads',
     "thread",
+
+    # whatis
+    "dbuf |head 1 |deref |member db_buf |whatis",
+    "whatis 0xffffa089407ca870",
+    "whatis 0xffffa0888c766000 0xffffa089407ca870",
+    "whatis 0xffff",
+    "whatis 0xf987kkbbh"
```
sdimitro, line 103: Great test cases!

sdb/commands/linux/whatis.py
```diff
@@ -0,0 +1,76 @@
...
+
+def lookup_in_kmem(obj: drgn.Object) -> None:
+    pfn = mm.virt_to_pfn(sdb.get_prog(), obj)
+    page = mm.pfn_to_page(pfn)
+    try:
+        cache = page.slab_cache
+        cache_nm = cache.name.string_().decode('utf-8')
+        print(f"{hex(int(obj))} is allocated from {cache_nm}")
+    except drgn.FaultError:
+        print(f"{hex(int(obj))} does not map to a kmem_cache")
```
sdimitro, line 36: I think we should do a couple of changes to this helper if we are to truly make it re-usable:
[1] Move it to `sdb/commands/linux/internal/slub_helpers.py`
[2] substitute the `mm` alias with the conventions of the above helper file (e.g. import `virt_to_pfn` and `pfn_to_page` at the top of that file and use them without `mm`).
[3] Change the functions signature:
```
def lookup_cache_by_address(obj: drgn.Object) -> Optional[drgn.Object]:
```
[4] Move the printing to the sdb command itself and have that helper method return the cache as a drgn.Object. If we get a `FaultError` because the page is not part of the cache we return `None` and the caller/command needs to handle it.

sdb/commands/linux/whatis.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class WhatIs(sdb.Command):
```
sdimitro, line 39: This is more of a question: Did you have any problems making this a pretty printer since it just prints?

sdb/commands/linux/whatis.py
```diff
@@ -0,0 +1,76 @@
...
+        Determine the kmem_cache for a given address:
+
+            sdb> whatis 0xfffffe06596e21b8
+            0xfffffe06596e21b8 is allocated from dmu_buf_impl_t
+
+        Determine the kmem_cache of address passed through a
+        pipe:
+
+            sdb> dbuf |head 1 |deref |member db_buf |whatis
+            0xffff8e804368ac80 is allocated from arc_buf_t
```
sdimitro, line 57: Thank you for writing the help messages this is very useful!

sdb/commands/linux/whatis.py
```diff
@@ -0,0 +1,76 @@
...
+        parser.add_argument("address", nargs="*", metavar="<address>")
+        return parser
+
+    def _call(self, objs: Iterable[drgn.Object]) -> None:
+        for obj in objs:
+            lookup_in_kmem(obj)
+        for addr in self.args.address:
+            try:
+                lookup_in_kmem(
+                    drgn.Object(sdb.get_prog(), "void *", int(addr, 16)))
```
sdimitro, line 74: We generally try to reduce the invocation of raw `drgn` methods in commands. Instead of using `drgn.Object(...)` you could use `create_object(...)` from `sdb.target.py`.

sdimitro (changes requested):
This is great and thank you for working on this!

Have you given any thoughts on how we can also check if the address provided is part of an independent zfs spl-cache? No worries if you haven't we can still go on with just linux-owned cache files. I was just curious.

PaulZ-98:
> This is great and thank you for working on this!
> 
> Have you given any thoughts on how we can also check if the address provided is part of an independent zfs spl-cache? No worries if you haven't we can still go on with just linux-owned cache files. I was just curious.

I had started down the road of building up a list of all the caches with their starting address and size, to see if the requested address fell within one of the ranges, and was thus from that kmem_cache.  I wonder if there is a better way to see if the address is in a specific zfs spa-cache instead of a linux cache.

sdb/commands/linux/whatis.py
PaulZ-98: I'm going to try to make it a pretty printer.

.github/scripts/install-drgn.sh
```diff
@@ -8,7 +8,6 @@ sudo apt update
 sudo apt install bison flex libelf-dev libdw-dev libomp5 libomp-dev
 
 git clone https://github.com/osandov/drgn.git
-
```
sdimitro, line 11: Unneeded whitespace change

sdimitro (approved):

PR #270, opened 2021-03-17
Add zfs_refcount command.

PaulZ-98:
## Pull request checklist

- [x] Build was run locally and any changes were pushed
- [x] Lint has passed locally and any fixes were made for failures

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: 192

## What is the new behavior?
sdb> spa | member spa_refcount | zfs_refcount
0xffffffffc1089360   dsl_sync_task_common 
0xffff91223d23a000    
0xffff911f9659f000    
0xffff91223f870800    
0xffff91223f870800    
0xffff911f9659c000    
0xffff912249c9d800    
0xffff912249c9d800    
0xffff911f9659c000    
0xffff91223caa6800    
0xffff91223caa6800    
0xffff911f9659c000    
0xffff91223caa4800    

- [ ] Yes
- [x] No

## Other information
Requires a crash dump with reference tracking enabled, so some of the testing infrastructure was refactored.


.github/scripts/install-drgn.sh
```diff
@@ -10,5 +10,6 @@ sudo apt install bison flex libelf-dev libdw-dev libomp5 libomp-dev
 git clone https://github.com/osandov/drgn.git
 
 cd drgn
+git checkout -b drgn_february 0a6aaaae5d31ed142448f8220e208d65478ec80d
```
sdimitro, line 13: Can we remove this line? If there is a problem with the github actions we fix the actions themselves. If there is a problem with drgn we fix drgn.

.github/scripts/download-dump-from-s3.sh
```diff
@@ -16,7 +16,11 @@
 #           └── vmlinux-5.0.0-36-generic
 #
 
+# use the default dir unless alternate dir passed as arg2 to script
 DATA_DIR="tests/integration/data"
+if [ ! -z "$2" ]; then
+	DATA_DIR=$2
+fi
```
sdimitro, line 23: I think instead of giving the option for which dump to use, we should just go ahead and test both dumps, no option needed. This means a lot of things, so here is a preliminary checklist of the things we need to do:
[1] This script can accept and download more than one dumps, then placed them into their own respective directories under the `DATA_DIR` (the comment above will also need to be changed along with the code)
[2] The test infrastructure should be made aware that multiple dumps can exist and always run all commands that we have for both crash dumps
[3] The saved regression output should be moved into crash-dump-specific regression output directories
[4] The github actions will need to be adjusted to download both crash dumps
[5] The wiki steps to run the tests manually will need to be updated on the wiki to include both dumps (https://github.com/delphix/sdb/wiki/Integration-Tests)

We should do the above over multiple pull requests:
[1st PR] The first pull request should make the changes to this download script to accept multiple dumps, and the respective changes to the test infrastructure code and regression output
[2nd PR] The second pull request will be adding the second crash dump and its own regression output
[3rd PR] The last pull request will be adding the new zfs_refcount command

Another cool thing would be for the new crash dumps themselves to keep a README file on their compressed archive highlighting why we added it (e.g. old crash dump X did not have refcounts enabled and we want to test that).

tests/integration/infra.py
```diff
@@ -24,116 +24,19 @@
...
-    test_mod_generic which has a list of commands in string
-    form called CMD_TABLE.
-    """
-    test_mod = import_module(f"tests.integration.test_{modname}_generic")
-    generate_output_for_commands(
-        test_mod.CMD_TABLE,  # type: ignore[attr-defined]
-        f"{TEST_OUTPUT_DIR}/{modname}")
-    print(f"Generated regression test output for {modname}...")
+PRIMARY = "primary"
+ALTERNATE = "alternate"
```
sdimitro, line 36: Instead of `PRIMARY` and `ALTERNATE` let's go ahead and have a list of crash dumps automatically detected by the target `DATA_DIR`. For example this would currently look like this with the two crash dumps that we have:
```
DATA_DIR/dump-<date 1>/{dump, mods, vmlinux, regression_output}
DATA_DIR/dump-<date 2>/{dump, mods, vmlinux, regression_output}
```

Keeping this structure as is makes it easy to add new crash dumps in the future.

tests/integration/infra.py
```diff
@@ -24,116 +24,19 @@
...
 
 
+@staticmethod
```
sdimitro, line 39: static methods make sense in class context context. You can move this method in the `Infra` class.

sdimitro (changes requested):
Hey @PaulZ-98, thank you for working on this!

There is a lot of things going on here, so let's break this into multiple commits/PRs.

sdb/commands/zfs/zfs_refcount.py
```diff
@@ -0,0 +1,56 @@
...
+
+# pylint: disable=missing-docstring
+
+from typing import Iterable
+import drgn
+import sdb
+from sdb.commands.spl.spl_list import SPLList
+
+
+class Zfs_Refcount(sdb.Locator, sdb.PrettyPrinter):
```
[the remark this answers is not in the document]
PaulZ-98: I changed it to just a pretty-printer.

class Zfs_Refcount(sdb.PrettyPrinter):

PaulZ-98:
I made a version with just the zfs_refcount command and standard test.

sdb/commands/zfs/zfs_refcount.py
```diff
@@ -0,0 +1,56 @@
...
+    names = ["zfs_refcount"]
+    input_type = "zfs_refcount_t *"
+    output_type = "zfs_refcount_t *"
+
+    @staticmethod
+    def print_ref(obj: drgn.Object):
+        ptr = int(obj.ref_holder)
+        c = sdb.create_object("char *", ptr)
+        s = c.string_().decode("utf-8")
```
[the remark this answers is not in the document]
PaulZ-98: Hi @ahrens I played with this a bit and did finally get an exception generated from the decode method.  I added a try/catch for it.

Issue #248, opened 2020-10-12
Add "ps" command; either new, or alias to "threads"


prakashsurya:
While looking at a kernel core dump, I wanted to look at all the processes that may have been running at the time of the crash. My first instinct was to run "ps", but that command doesn't exist. I eventually found the "threads" command, which is exactly what I needed, but took me a little while longer to find the command.

Feel free to close if this goes against the design of "threads", but it may be more user-friendly (at least for me) to have a "ps" command since that is the first thing that I tried.

sdimitro:
I think having `ps` is a good idea for people that don't know about `threads` to start from somewhere - it could even be a layer build on top `threads` (instead of just an alias) that makes it more process-centric and group tasks that have the same PID together.

PR #274, opened 2021-04-06
Add the ps command to sdb.

tony-zfs:
## Pull request checklist

- [X ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X ] Build was run locally and any changes were pushed
- [X ] Lint has passed locally and any fixes were made for failures

## Pull request type

 

 

- [ ] Bugfix
- [ X] Feature
- [ ] Build related changes

Issue Number: 248
https://github.com/delphix/sdb/issues/248

## What is the new behavior?

- Supports limited version of ps
-
-

- [ ] Yes
- [X ] No


sdb/commands/linux/ps.py
```diff
@@ -0,0 +1,180 @@
...
+        #
+        # The command line information is contained in the user address
+        # space of each thread, rather than in the kernel's address
+        # space. Thus, often, it may not be possible to retreive the
+        # thread's command line; e.g. when reading from a core dump.
+        #
+        return ""
+
+
+class Ps(sdb.Locator, sdb.PrettyPrinter):
```
sdimitro, line 55: A lot of code in this class overlaps with `threads.py`. Instead of duplicating code in both places did you try the following:
[1] Extend the `threads` command to handle the extra columns that you added (e.g. `uid`) and also the column logic (e.g `-o` , `-h`, etc..)
[2] Make the `ps` command use the `threads` command under the hood and keep the arguments that make it look like the linux userland ps command (e.g. `-p`, `--ppid`, etc..).

If you want an example of layering a command on top of another command checkout the `spa` command. It is using  the `avl` walker for its Locator/no-input method, and `vdev` for its pretty-printer.

As for generic use of the `Table` class and how to implement the sort-by-column functionality you can look at either the `slubs` or `spl_kmem_caches`.

sdb/commands/linux/ps.py
```diff
@@ -0,0 +1,180 @@
...
+        parser = super()._init_parser(name)
+        parser.add_argument('-e', '--every', action='store_true', \
+                help="Select all processes. Identical to -A")
+        parser.add_argument('-A', '--all', action='store_true', \
+                help="Select all processes. Identical to -e")
+        parser.add_argument('-C', '--C', type=str, \
+                help="Print only the process IDs of a command")
+        parser.add_argument('-x', '--x', action='store_true', \
+                help="Show PID, TIME, CMD")
+        parser.add_argument('-h', action='store_true', \
```
sdimitro, line 116: Unfortunately `argparse` doesn't allow us to use `-h` for anything other than the help message. Fortunately for us looking at `man ps` I see that that to skip the headers only the long form argument exists `--no-headers` and `--no-heading`.

sdb/commands/linux/ps.py
```diff
@@ -0,0 +1,180 @@
...
+        0xffff99950a430000 1000 4284 0:00:00 0:00:00 4118  HTML5 Parser
+        0xffff99950a431680 1000 4157 0:00:12 0:00:08 1218  localStorage DBa
+        ...
+    """
+
+    names = ["ps"]
+    input_type = "struct task_struct *"
+    output_type = "struct task_struct *"
+
+    FIELDS: Dict[str, Callable[[drgn.Object], Union[str, int]]] = {
```
sdimitro, line 86: Not that it matters if we go ahead with layering the `ps` command on top of `threads` but I don't see `task` mentioned here. What if a user wants to specify it with the `-o` option?

sdb/commands/linux/ps.py
```diff
@@ -0,0 +1,180 @@
...
+                help="Select all processes. Identical to -e")
+        parser.add_argument('-C', '--C', type=str, \
+                help="Print only the process IDs of a command")
+        parser.add_argument('-x', '--x', action='store_true', \
+                help="Show PID, TIME, CMD")
+        parser.add_argument('-h', action='store_true', \
+                help="Show the output without headers.")
+        parser.add_argument('-p', '--p', type=str, \
+                help="Select by process ID. Identical to --pid.")
+        parser.add_argument('-ppid', '--ppid', type=str, \
```
sdimitro, line 120: Having the long-form and the short-form argument look the same doesn't look right. Looking at `man ps` it seems like only the long-form is supported `--ppid`. If you really want a short-form I don't think `-P` is taken, so we could use that.

In addition, like `-p/--pid` the type of argument can be a list of integers.

sdb/commands/linux/ps.py
```diff
@@ -0,0 +1,180 @@
...
+                help="Select all processes. Identical to -A")
+        parser.add_argument('-A', '--all', action='store_true', \
+                help="Select all processes. Identical to -e")
+        parser.add_argument('-C', '--C', type=str, \
+                help="Print only the process IDs of a command")
+        parser.add_argument('-x', '--x', action='store_true', \
+                help="Show PID, TIME, CMD")
+        parser.add_argument('-h', action='store_true', \
+                help="Show the output without headers.")
+        parser.add_argument('-p', '--p', type=str, \
```
sdimitro, line 118: Looking at `man ps` I think the correct long-form for this is `--pid`.

sdb/commands/linux/ps.py
```diff
@@ -0,0 +1,180 @@
...
+            fields_option_o = [x for x in fields if x in fields_option_o]
+            fields = fields_option_o
+        else:
+            fields = fields_option_default
+
+        table = Table(fields, None, {"PID": str})
+
+        cmds = self.args.C.split(',') if self.args.C else []
+        pids = [int(x) for x in self.args.p.split(',')] if self.args.p else []
+        ppids = [int(x) for x in self.args.ppid.split(',')
```
sdimitro, line 159: what happens here when we get faulty input? (e.g. "asdf, asdfasdf")

If the result is that we blow up maybe it may be easier to switch from the `ps --pid 1,2,3` to `ps --pid 1 2 3` notation (both are supported from the userland `ps` utility in Linux from what I can tell). This would mean that we'd need to change the argument type to something like `type=int, nargs="+",`.

tests/integration/test_linux_generic.py
```diff
@@ -89,6 +89,15 @@
...
     "echo 0xffffa089669edc00 | stack",
 
+    # ps
+    "ps",
+    "ps -C bash",
+    "ps -p 4275",
+    "ps -e",
+    "ps -x",
+    "ps -o pid,ppid",
+    "ps -A",
```
sdimitro, line 99: Thanks for extending the test-suite!

sdb/commands/stacks.py
```diff
@@ -152,7 +152,7 @@ def __init__(self,
         self.match_state = ""
 
     @classmethod
-    def _init_parser(cls, name: str) -> argparse.ArgumentParser:
+    def _init_parser(cls, name: str) -> argparse.ArgumentParser():
```
sdimitro, line 155: Pressumably `mypy` bothered you about this change. We've found it to be extremely inconsistent with false-positives being introduced at every new release thus we've frozed our version of mypy in our Github Actions to `0.730` - https://github.com/delphix/sdb/blob/master/.github/workflows/main.yml#L100 . You may want to do the same in your python environment for sdb locally.

sdimitro (changes requested):
Hey @tony-zfs thank you for opening the PR!

From what I see the failures in the test-suite are from our frozed version of `mypy` and the use of `-h` option from `argparse`. I have some more details on both bundled together with the rest of my feedback.

sebroy:
I would love to see sample output of the new subcommand as part of the PR.

sdb/commands/stacks.py
tony-zfs: Thanks - it sure did!

sdb/commands/linux/ps.py
tony-zfs: Nice catch - I'll update.

sdb/commands/linux/ps.py
tony-zfs: Updated. Thanks for the advice.

sdb/commands/linux/ps.py
```diff
@@ -0,0 +1,180 @@
...
+import drgn
+from drgn.helpers.linux.pid import for_each_task
+from drgn.helpers.linux.mm import cmdline
+
+import sdb
+from sdb.commands.internal.table import Table
+from sdb.commands.stacks import Stacks
+
+
+def _cmdline(obj: drgn.Object) -> str:
```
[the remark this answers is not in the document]
tony-zfs: removed.

sdb/commands/linux/ps.py
tony-zfs: updated.

sdb/commands/linux/ps.py
tony-zfs: updated.

sdb/commands/linux/ps.py
tony-zfs: updated.

sdb/commands/linux/ps.py
tony-zfs: fixed

tony-zfs:
Updated. Is there anything else needed to get this to be accepted?

PaulZ-98:
> I would love to see sample output of the new subcommand as part of the PR.

If you drill into the commit, there are sample outputs used in the automated testing of the new command.  Look for ` tests/integration/data/regression_output/linux/ps -e` and the like.

PR #275, opened 2021-04-08
Replace current PR template on Github

sdimitro:
= Problem
The default PR template that we inherit from the Delphix Github
organization is too generic and some of its fields are not applicable
to SDB.

= Solution
This patch introduces a new template that is more suitable for our
needs.

= Notes To Reviewers
The choice of using `=` instead of `#` (Markdown Header) for
the template sections is for the case that the submitter decides
to copy paste the template in their commit description in git.
The `#` character is used for comments in git descriptions so
these section headers would disappear once the commit description
is written.

= Testing
I've opened the following test PR where I verify that I can still see
the comments in the PR description when creating or editing it:
https://github.com/sdimitro/sdb/pull/4

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,35 @@
+[//]: # (This PR template outlines different sections that you can fill so reviewers have an easier time understanding your review.)
+[//]: # (Using this template is strongly encouraged but not a hard requirement for getting your PR merged.)
+[//]: # (Using this template in your commit description is also encouraged.)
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):

ahrens:
The rendering on the webpage is unfortunate (`=` instead of large/bold), but if you think the git comment copy/paste is more important then I'm OK with it.

ahrens (approved):

PR #278, opened 2021-04-29
Fix Current Pylint Errors

sdimitro:
= Before The Patch

We'd get the following error from our `pylint` action:
```
************* Module sdb.pipeline
sdb/pipeline.py:132:8: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
************* Module sdb.commands.zfs.histograms
sdb/commands/zfs/histograms.py:136:8: R1731: Consider using 'max_count = max(max_count, HISTOGRAM_WIDTH_MAX)' instead of unnecessary if block (consider-using-max-builtin)
```

= After the Patch

Pylint runs fine with no errors

### DIFF
--- a/sdb/commands/zfs/histograms.py
+++ b/sdb/commands/zfs/histograms.py
@@ -133,8 +133,7 @@ class ZFSHistogram(sdb.Command):
                 max_count = count
 
         HISTOGRAM_WIDTH_MAX = 40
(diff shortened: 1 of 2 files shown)


mmaybee (approved):
Sigh, pylint seems a bit pedantic to me sometimes.

prakashsurya (approved):

PR #285, opened 2022-05-17
determine the vmlinux file from the crash dump

### LINKED ISSUE #41: how to determine which "vmlinux" to use when working with kernel dumps?
Currently, it's possible to debug a kernel crash dump with SDB by running something like this:
```
$ sdb /usr/lib/debug/boot/vmlinux-4.15.0-65-generic /var/crash/201910142024/dump.201910142024 -s /lib/modules/4.15.0-65-generic
```
The problem with this approach, though, is there's no guarantee that the crash dump (`dump.201910142024`) should be used with the "vmlinux" file (`vmlinux-4.15.0-65-generic`) passed to SDB.

For example, if multiple kernels are installed in `/usr/lib/debug/boot` (e.g. because multiple kernel packages are installed), how should the user determine which "vmlinux" to use when starting SDB? Additionally, what could happen if the wrong "vmlinux" file was specified? My guess is SDB my break in silent and subtle ways (i.e. it may show the wrong data, without outright failing).

I think we should improve the usability of SDB and harden it against incorrect usage, by not requiring the user to pass in the "vmlinux". For example, the dump contains some information about the kernel version that it needs:
```
$ sudo dd if=dump.201910142024 bs=1M count=1 2>/dev/null | strings | grep OSRELEASE
OSRELEASE=4.15.0-65-generic
```
Thus, we could modify SDB to look for this `OSRELEASE` information in the dump, and then automatically search the root filesystem for the "vmlinux" file that corresponds to that dump. Then, the interface for SDB could more simply be:
```
$ sdb /var/crash/201910142024/dump.201910142024
```
Which would prevent any possibility for the user to incorrectly pass in the wrong "vmlinux" file.

PaulZ-98:
= Problem

sdb does not automatically determine the vmlinux file from the crash dump
https://github.com/delphix/sdb/issues/41

= Solution

When a crash dump is supplied but the vmlinux file is not supplied on the command line, parse the crash dump to find the OS version, then scan /usr/lib/debug and /root for the vmlinux file matching that version.   This follows the solution proposed by @prakashsurya.  Other options require changes outside of sdb/drgn.

root@ub-20:~/sdb_detect/sdb# sdb -s /root/objs /root/dump-orig/dump.202102031354
sdb using ['/root/dumps/vmlinux-5.8.0-41-generic'] auto-detected from core file
sdb>

[//]: # (= Future work)

[//]: # (A description of what follow up work is explicitly not being done in this change.)
[//]: # (Creating new issues for that future work in the repo tracker and mentioning them here may work as well.)

[//]: # (= Testing)

Tested with a core referencing a vmlinux that can be found in /root or /usr/lib/debug, and with one that is not located in /root or /usr/lib/debug.

[//]: # (= Github Issue Tracker Automation)

Closes #41


PaulZ-98:
Let me try this again.  It picks up too many other Datto changes.

PR #286, opened 2022-05-17
sdb should detect vmlinux version from crash dump

### LINKED ISSUE #41: how to determine which "vmlinux" to use when working with kernel dumps?

### LINKED ISSUE #56: sdb requires sudo just for the history file to be accessed

PaulZ-98:
Signed-off-by: Paul Zuchowski <pzuchowski@datto.com>

[//]: # (This PR template outlines different sections that you can fill so reviewers have an easier time understanding your review.)
[//]: # (Using this template is strongly encouraged but not a hard requirement for getting your PR merged.)
[//]: # (Using this template in your commit description is also encouraged.)

= Problem

sdb does not automatically determine the vmlinux file from the crash dump
https://github.com/delphix/sdb/issues/41

= Solution

When a crash dump is supplied but the vmlinux file is not supplied on the command line, parse the crash dump to find the OS version, then scan /usr/lib/debug and /root for the vmlinux file matching that version. This follows the solution proposed by @prakashsurya. Other options require changes outside of sdb/drgn.

root@ub-20:~/sdb_detect/sdb# sdb -s /root/objs /root/dump-orig/dump.202102031354
sdb using ['/root/dumps/vmlinux-5.8.0-41-generic'] auto-detected from core file
sdb>

[//]: # (Extra Possible Sections:)

[//]: # (= Notes To Reviewers)

[//]: # (Any extra information a reviewer may need to know before reviewing your change.)
[//]: # (For example here you might want to describe which files should be looked at first or which files.)

[//]: # (= Future work)

[//]: # (A description of what follow up work is explicitly not being done in this change.)
[//]: # (Creating new issues for that future work in the repo tracker and mentioning them here may work as well.)

Tested with a core referencing a vmlinux that can be found in /root or /usr/lib/debug, and with one that is not located in /root or /usr/lib/debug.

[//]: # (= Github Issue Tracker Automation)

[//]: # (If your patch is fixing an issue filled in the repo's issue tracker mention this here. Example: "Closes #56")
Closes #41

### DIFF
--- a/sdb/internal/cli.py
+++ b/sdb/internal/cli.py
@@ -21,8 +21,9 @@ like the entry point, command line interface, etc...
 import argparse
 import os
 import sys
+import string
 
-from typing import List
+from typing import Iterable, List
 
 import drgn
 import sdb
@@ -117,22 +118,31 @@ def parse_arguments() -> argparse.Namespace:
         parser.error(
             "cannot specify an object file while also specifying --pid")
 
-    #
-    # We currently cannot handle object files without cores.
-    #
-    if args.object and not args.core:
-        parser.error("raw object file target is not supported yet")
-
     return args
 
 
-def load_debug_info(prog: drgn.Program, dpaths: List[str]) -> None:
+def load_debug_info(prog: drgn.Program,
+                    dpaths: List[str],
+                    vml: str = "") -> None:
     """
     Iterates over all the paths provided (`dpaths`) and attempts
     to load any debug information it finds. If the path provided
     is a directory, the whole directory is traversed in search
     of debug info.
     """
+    if vml:
+        found = False
+        for target in ['/usr/lib/debug', '/root']:
+            vmlinux = []
+            for root, _dirs, files in os.walk(target):
+                if vml in files:
+                    found = True
+                    vmlinux.append(os.sep.join([root, vml]))
+                    print(f"sdb using {vmlinux} auto-detected from core file")
+                    prog.load_debug_info(vmlinux)
+                    break
+            if found:
+                break
     for path in dpaths:
         if os.path.isfile(path):
             prog.load_debug_info([path])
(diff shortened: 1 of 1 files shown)


sdb/internal/cli.py
```diff
@@ -147,26 +157,88 @@ def load_debug_info(prog: drgn.Program, dpaths: List[str]) -> None:
...
+
+def is_core(filename: str) -> bool:
+    """
+    Parse beginning of the given file to see if it starts with 'KDUMP'
+    """
+    match = "KDUMP"
+    for s in strings(filename, len(match), 64):
+        if s.startswith(match):
+            return True
+    return False
```
sdimitro, line 206: This is a bit hacky. How about we use the Python API of https://github.com/ptesarik/libkdumpfile to first check if it is a valid kdump file and then use query its build-id parameter/property to get our matching?

sdimitro (changes requested):

sdb/internal/cli.py
PaulZ-98: Can you give me some hints on which API calls I could use?  I'm not able to find a clean way to do this with libkdumpfile.

PR #291, opened 2022-08-01
Fix SDB Regressions

sdimitro:
Make all the Github CI checks to pass again.

Most issues fixed are related to code formatting and linting.

Besides those, the rest of the regression failures have to do
with drgn returning duplicate stack frame entries for inlined
functions. We don't attempt to do anything about those
failures for now other than recreating our reference/expected
command ouput for the `stacks` command regression tests
to match what drgn returns (tackling this could be tricky,
especially in situations where a stacktrace involves recursion).

Finally we update the OS of the CI runs to Ubuntu 20.04 and
"raise" our officially minimal supported python version to
python 3.8. Older versions like 3.6 and 3.7 should still work
but maintaining CI runs of them can get annoying over time
as `pylint` of 3.8 may raise issues not raised in `pylint` of
python 3.7 and vice versa.

### DIFF
--- a/.github/scripts/install-libkdumpfile.sh
+++ b/.github/scripts/install-libkdumpfile.sh
@@ -8,7 +8,7 @@
 #
 sudo apt update
 sudo apt install autoconf automake liblzo2-dev libsnappy1v5 libtool pkg-config zlib1g-dev
-sudo apt install python3.6-dev python3.7-dev python3.8-dev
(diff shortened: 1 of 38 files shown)


ahrens (approved):

prakashsurya (approved):
I think you'll need to land this directly to `6.0/stage` now.

PR #309, opened 2023-01-30
DLPX-84437 Print panicked thread of crash dump on SDB

sdimitro:
= Motivation

The `stacks` command can be overwhelming for newcomers that need to
figure out which thread caused the debugging target to panic. An
alternative would be `dmesg`, but that command also has a lot of
extraneous output.

= This Patch

Creates a new SDB command that yields the panicked thread to subsequent
commands or just pretty-prints it in the style of the `stacks` command.

= Example Output

```
sdb> crashed_thread
0xffff8f15d7333d00 RUNNING               1
	      __crash_kexec+0x9d
	      __crash_kexec+0x9d
	      panic+0x11d
	      0xffffffff9020b375+0x0
	      __handle_sysrq.cold+0x48
	      write_sysrq_trigger+0x28
	      proc_reg_write+0x43
	      __vfs_write+0x1b
	      vfs_write+0xb9
	      vfs_write+0xb9
	      ksys_write+0x67
	      __x64_sys_write+0x1a
	      __x64_sys_write+0x1a
	      __x64_sys_write+0x1a
	      do_syscall_64+0x57
	      entry_SYSCALL_64+0x94
```

Error case A: Live-Target:
```
$ sudo sdb
sdb> crashed_thread
sdb: crashed_thread: command only works for core/crash dumps
```

Error case B: Pretty-print bogus thread/value:
```
sdb> stacks | head 1 | crashed_thread
sdb: crashed_thread: can only pretty print the crashed thread
sdb> echo 0x0 | crashed_thread
sdb: crashed_thread: can only pretty print the crashed thread
```

### DIFF
--- a/sdb/commands/stacks.py
+++ b/sdb/commands/stacks.py
@@ -400,3 +400,54 @@ class Stacks(sdb.Locator, sdb.PrettyPrinter):
     def no_input(self) -> Iterable[drgn.Object]:
         self.validate_context()
         yield from filter(self.match_stack, for_each_task(sdb.get_prog()))
+
+
+class CrashedThread(sdb.Locator, sdb.PrettyPrinter):
+    """
+    Print the crashed thread. Only works for crash dumps and core dumps.
+
+    EXAMPLES
+        sdb> crashed_thread
+        TASK_STRUCT        STATE             COUNT
+        ==========================================
+        0xffff8f15d7333d00 RUNNING               1
+                          __crash_kexec+0x9d
+                          __crash_kexec+0x9d
+                          panic+0x11d
+                          0xffffffff9020b375+0x0
+                          __handle_sysrq.cold+0x48
+                          write_sysrq_trigger+0x28
+                          proc_reg_write+0x43
+                          __vfs_write+0x1b
+                          vfs_write+0xb9
+                          vfs_write+0xb9
+                          ksys_write+0x67
+                          __x64_sys_write+0x1a
+                          __x64_sys_write+0x1a
+                          __x64_sys_write+0x1a
+                          do_syscall_64+0x57
+                          entry_SYSCALL_64+0x94
+    """
+
+    names = ["crashed_thread", "panic_stack", "panic_thread"]
+    input_type = "struct task_struct *"
+    output_type = "struct task_struct *"
+
+    def validate_context(self) -> None:
+        if sdb.get_target_flags() & drgn.ProgramFlags.IS_LIVE:
+            raise sdb.CommandError(self.name,
+                                   "command only works for core/crash dumps")
+
+    def pretty_print(self, objs: Iterable[drgn.Object]) -> None:
+        self.validate_context()
+        thread_obj = sdb.get_prog().crashed_thread().object
+        stacks_obj = Stacks()
+        for obj in objs:
+            if obj.value_() != thread_obj.value_():
+                raise sdb.CommandError(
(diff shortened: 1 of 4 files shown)


mmaybee (approved):
Nice. Is there any precedent for the command name? `crashed_thread` is OK, but seems slightly awkward to me... how about `panic_stack` or `panic_thread`?

sdimitro:
@mmaybee both sound good, I can add them as command aliases

sdb/commands/stacks.py
```diff
@@ -400,3 +400,54 @@ def pretty_print(self, objs: Iterable[drgn.Object]) -> None:
...
+                          ksys_write+0x67
+                          __x64_sys_write+0x1a
+                          __x64_sys_write+0x1a
+                          __x64_sys_write+0x1a
+                          do_syscall_64+0x57
+                          entry_SYSCALL_64+0x94
+    """
+
+    names = ["crashed_thread", "panic_stack", "panic_thread"]
+    input_type = "struct task_struct *"
```
ahrens, line 433: It looks like we don't really intend to pipe things to this command.  Could we instead say that it doesn't take any input, and then have the no_input() method do the printing, if the output is not a pipe?  (not sure if that would actually be cleaner with the current infrastructure)

Alternatively, would it make sense to decompose this into 2 commands:  one that prints the stack trace given a `task_struct*`, and one that outputs the current (crashed) `task_struct*`?  e.g. `sdb> current_task | task_backtrace`.  Although maybe that's too wordy for a common command, gdb has `bt`

ahrens (approved):

sdb/commands/stacks.py
sdimitro: I did consider your first suggestion where we basically make that command a subclass of `Locator` but not `PrettyPrinter` and check `self.islast` to either print or yield the object.

The main pro is that we don't need to write the extra code that I have now for the pretty printer.
The main con is that conceptually `no_input()` functions of `Locators` generally just yield Objects and don't print anything - we could actually workaround that by making the function a standard `Command` instead of a `Locator`.

Personally I thought that making this both a Locator and a PrettyPrinter may be nice for two reasons:
* If its at the end of a pipeline it can be used as an ad-hoc predicate - "is this a thread the crashed thread?" (e.g. `echo <some zio addr> | member io_waiter | crashed_thread)
* The command is to be used by itself or at the beginning of the pipeline it works fine, but if it is in the middle or the end of a pipeline it will give an error if you give it anything other than the crashed thread. If we were to make it a `Locator/Command` this behavior would change to dropping everything listed before in the pipeline. This is probably not a big deal and a scenario that would almost never happen but it did feel awkward to me. Admittedly there are other commands like that (e.g. `spa | exit` or `spa | history`) - but they just seem more related to the sdb session and not specific to the target we analyze.

----

Decomposing the command to two might make sense but it doesn't necessarily deal with the above point of what happens when we insert this in the middle of an existing pipeline. I also want to avoid the name `current_task` for now because I don't think we've decided if SDB should have a task-aware context in the future or ever at all. Still `crashed_thread`/`panic_stack`  could just be Locators but as you said the result may be too wordy. A way around this could be that we created a 3rd command called `bt` that always executes `crashed_thread | stacks` under the covers.

Thoughts?

ahrens: That makes sense.  I think it would be fine to keep it as-is and we can revisit later, depending on our experience using it.

PR #312, opened 2023-02-10
Update btree walker to be aware of btree optimizations

pcd1193182:
[//]: # (This PR template outlines different sections that you can fill so reviewers have an easier time understanding your review.)
[//]: # (Using this template is strongly encouraged but not a hard requirement for getting your PR merged.)
[//]: # (Using this template in your commit description is also encouraged.)

= Problem
ZFS's btrees were recently optimized to improve performance. One of the changes modified the in memory structure by adding a first-element offset for leaf nodes, allowing a reduction in memmove calls and sizes.

= Solution
Update the btree walker to be aware of this new behavior

### DIFF
--- a/sdb/commands/zfs/btree.py
+++ b/sdb/commands/zfs/btree.py
@@ -88,7 +88,7 @@ class Btree(sdb.Walker):
             # generate each object in the leaf elements
             leaf = drgn.cast('struct zfs_btree_leaf *', node)
             for i in range(count):
-                yield self._val(leaf.btl_elems, i)
+                yield self._val(leaf.btl_elems, i + node.bth_first)
 
     def walk(self, obj: drgn.Object) -> Iterable[drgn.Object]:
         self.elem_size = obj.bt_elem_size


ahrens (approved):

sdb/commands/zfs/btree.py
```diff
@@ -88,7 +88,7 @@ def _helper(self, node: drgn.Object) -> Iterable[drgn.Object]:
             # generate each object in the leaf elements
             leaf = drgn.cast('struct zfs_btree_leaf *', node)
             for i in range(count):
-                yield self._val(leaf.btl_elems, i)
+                yield self._val(leaf.btl_elems, i + node.bth_first)
```
sdimitro, line 91: This is breaking old kernels can we maybe do something like this:
```
        try:
+          bth_first = node.bth_first
            recurse = int(bth_first) == -1
        except AttributeError:
            bth_first = 0
            recurse = node.bth_core
        count = node.bth_count
        if recurse:
            # alterate recursive descent on the children and generating core objects
            core = drgn.cast('struct zfs_btree_core *', node)
            for i in range(count):
                yield from self._helper(core.btc_children[i])
                yield self._val(core.btc_elems, i)
            # descend the final, far-right child node
            yield from self._helper(core.btc_children[count])
        else:
            # generate each object in the leaf elements
            leaf = drgn.cast('struct zfs_btree_leaf *', node)
            for i in range(count):
-                yield self._val(leaf.btl_elems, i)
+                yield self._val(leaf.btl_elems, i + bth_first)
```

sdimitro (changes requested):
This is an omission on my end when I first tried to fix this, so good catch and thanks for opening a PR!

The code itself looks good with new kernels but its breaking testing and github actions of the older ones.  I gave a suggestion on how to fix it.

PR #319, opened 2023-03-15
Add module system to prevent commands from being loaded when they aren't useful

### LINKED ISSUE #303: Need Command Module System
## Motivation

sdb can be used for userland core dumps but most current commands assume that we are in the kernel. E.g. `stacks` blows up in userland core dumps and other commands like `slabs` make no sense.

## What we need

SDB should detect:
* Whether its ran against a kernel crash dump or a userland core dump and load the appropriate SDB commands (e.g. userland `stacks` vs kernel `stacks`).
* If kernel crash dump - which kernel modules are loaded and load the appropriate SDB commands (e.g. zfs, nfs, etc..)
* If kernel crash dump - which config  parameters are set and load the appropriate SDB commands (e.g slab vs slub)
* If userland core - what's the main executable? (e.g. load zfs commands if ztest)
* If userland core - which libraries are loaded? (e.g. load zfs commands for libzpool)

pcd1193182:
= Problem
See #303 

= Solution
See #303. Specifically, the load_on field is added to every command, which allows a command to specify a list of runtimes in which it should be loaded. Until drgn supports providing a list of loaded modules and libraries, we choose to load all module-specific commands whenever we're in the kernel, and all library-specific ones whenever we're in userland. There are also a couple of small changes to prevent errors from being thrown when we try to interact with types that aren't loaded because their library/module isn't present.

= Testing
Ran against a running kernel and userland core files, verified that appropriate commands appeared in each and behaved as expected.

Closes #303

### DIFF
--- a/sdb/commands/spl/avl.py
+++ b/sdb/commands/spl/avl.py
@@ -27,7 +27,7 @@ class Avl(sdb.Walker):
 
     names = ["avl"]
     input_type = "avl_tree_t *"
-    load_on = [sdb.Module("zfs")]
+    load_on = [sdb.Module("spl")]
 
     def _helper(self, node: drgn.Object, offset: int) -> Iterable[drgn.Object]:
         if sdb.is_null(node):
--- a/sdb/commands/spl/avl.py
+++ b/sdb/commands/spl/avl.py
@@ -27,6 +27,7 @@ class Avl(sdb.Walker):
 
     names = ["avl"]
     input_type = "avl_tree_t *"
+    load_on = [sdb.Module("zfs")]
 
     def _helper(self, node: drgn.Object, offset: int) -> Iterable[drgn.Object]:
         if sdb.is_null(node):
--- a/sdb/internal/cli.py
+++ b/sdb/internal/cli.py
@@ -214,6 +214,8 @@ def main() -> None:
     except PermissionError as err:
         print("sdb: " + str(err))
         return
+    sdb.target.set_prog(prog)
+    sdb.register_commands()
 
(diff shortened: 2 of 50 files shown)


sdb/commands/spl/avl.py
```diff
@@ -27,6 +27,7 @@ class Avl(sdb.Walker):
 
     names = ["avl"]
     input_type = "avl_tree_t *"
+    load_on = [sdb.Module("zfs")]
```
sdimitro: strictly speaking all of the commands within the `spl` repo are part of the `spl` module, bug given that the `spl` module is not really used for anything else, and the `zfs` module cannot really work without `zfs` I'm fine with this left as-is

sdb/target.py
```diff
@@ -116,7 +116,11 @@ def type_canonicalize_name(type_name: str) -> str:
     """
     Return the "canonical name" of this type name.  See type_canonicalize().
     """
-    return type_canonical_name(prog.type(type_name))
+    # This is a workaround while we don't have module/library lookup in drgn.
```
sdimitro, line 120: Can you create an issue about this in GIthub and link it here?

sdb/internal/cli.py
```diff
@@ -214,6 +214,8 @@ def main() -> None:
     except PermissionError as err:
         print("sdb: " + str(err))
         return
+    sdb.target.set_prog(prog)
+    sdb.register_commands()
```
sdimitro, line 218: I'm a big fan of this!

sdimitro (approved):
LGTM!

prakashsurya (approved):

PR #281, opened 2021-08-24
Add blkptr command to sdb

PaulZ-98:
= Problem

sdb does not interpret the bitfields within ZFS block pointers in order to generate a proper display

= Solution

Implement the blkptr command in sdb.  It might have been possible to include some of the ZFS C code into python for this.  However it seemed the most straightforward to just port the functionality into python.

[//]: # (= Notes To Reviewers)
This was mostly porting C macros, but pylint guided the coding to have some object oriented content.

[//]: # (If your patch is fixing an issue filled in the repo's issue tracker mention this here. Example: "Closes #56")
Closes #27

PR #320, opened 2023-03-17
Add blkptr command to sdb

grwilson:
= Problem

This leverages the original work by @PaulZ-98 from PR https://github.com/delphix/sdb/pull/281. This add the `blkptr` command to sdb . 

= Solution

I tried to structure the code so that it is similar to the C code so that it's easier to enhance in the future. All of the `BP_*` macros that are used by ZFS were also added to make future enhancements easier.

### DIFF
--- a/sdb/commands/zfs/internal/__init__.py
+++ b/sdb/commands/zfs/internal/__init__.py
@@ -69,6 +69,10 @@ def BF64_GET(x: drgn.Object, low: int, length: int) -> int:
     return BF64_DECODE(x, low, length)
 
 
+def BF64_GET_SB(x: int, low: int, length: int, shift: int, bias: int) -> int:
+    return (BF64_GET(x, low, length) + bias) << shift
+
+
 def WEIGHT_IS_SPACEBASED(weight: int) -> bool:
     return weight == 0 or (BF64_GET(weight, 60, 1) != 0)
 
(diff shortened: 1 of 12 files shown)


prakashsurya (approved):

sdb/commands/zfs/internal/__init__.py
```diff
@@ -81,6 +85,188 @@ def WEIGHT_GET_COUNT(weight: int) -> int:
...
+    return BF64_GET(bp.blk_prop, 63, 1)
+
+
+def BP_GET_LAYER(bp: drgn.Object) -> int:
+    if BP_IS_EMBEDDED(bp):
+        if sdb.get_type('blkptr_t').has_member('blk_logical_birth'):
+            return BF64_GET(bp.blk_logical_birth, 56, 8)
+        return BF64_GET(bp.blk_birth, 56, 8)
+    if sdb.get_type('blkptr_t').has_member('blk_physical_birth'):
+        return BF64_GET(bp.blk_physical_birth, 56, 8)
```
ahrens: the layer info might be in the logical birth word, or it might be in the physical birth word?

tests/integration/test_zfs_generic.py
```diff
@@ -29,6 +29,10 @@
     # arc
     "arc",
 
+    # blkptr
+    "spa | head 1 | deref |member spa_uberblock | member ub_rootbp | blkptr",
+    "dbuf | head 1 | member db_blkptr | blkptr",
```
ahrens, line 34: adding these tests is great.  Don't you need to also add the expected output files?

sdimitro:
@grwilson I'm still in the middle of writing the Dev Resources for how to generate the test output so I just quickly generated it for you in this commit:
https://github.com/delphix/sdb/commit/b9110366735e112a5dfc2fc27d9630a209fb6971

In the above commit I also fixed a bug where you don't have the `load_on` global in your blkptr command - this is needed due to Paul D's recent changes. Also 2 of Paul Z's tests are no longer applicable so I removed them.

Finally can you please rebase your PR over the latest 6.0/stage branch?

sdb/commands/zfs/internal/__init__.py
grwilson: The layer is in both birth words if they exist (i.e. embedded blocks don't have physical birth). This is really just for consistency but I can change this macro to always look at the logical birth word since that should always exist.

ahrens (approved):

PR #321, opened 2023-03-20
Add userland stacks and threads commands

pcd1193182:
Now that the stacks and threads commands have been properly separated into the kernel module, we can write a userland implementation of these commands.  These implementations are heavily based on the originals, but using the generic drgn API that doesn't assume it can access the thread's object parameter. As a result, we have to lose the `module` and `state` filters on the stacks.

### DIFF
--- a/sdb/commands/ustacks.py
+++ b/sdb/commands/ustacks.py
@@ -45,42 +45,64 @@ class UserStacks(sdb.Locator, sdb.PrettyPrinter):
         The command returns all thread IDs that matched the
         filter.
 
+        The command will not produce output on running userland processes,
+        because the process is not stopped while being analyzed.
+
     EXAMPLES
         Print the call stacks for all threads
 
             sdb> stacks
-            TASK_STRUCT        STATE             COUNT
-            ==========================================
-            0xffff9521bb3c3b80 IDLE                394
-                              __schedule+0x24e
-                              schedule+0x2c
-                              worker_thread+0xba
-                              kthread+0x121
-                              ret_from_fork+0x35
-
-            0xffff9521bb3cbb80 INTERRUPTIBLE       384
-                              __schedule+0x24e
-                              schedule+0x2c
-                              smpboot_thread_fn+0x166
-                              kthread+0x121
-                              ret_from_fork+0x35
+            TID         COUNT
+            ==========================================================
+            ...
+            125023          1
+                              0x7f2fd24c1170+0x0
+                              spa_open_common+0x1ac
+                              dsl_sync_task_common+0x65
+                              dsl_sync_task_sig+0x13
+                              zcp_eval+0x70f
+                              dsl_destroy_snapshots_nvl+0x12b
+                              dsl_destroy_snapshot+0x3f
+                              ztest_objset_destroy_cb+0xb7
+                              dmu_objset_find_impl+0x481
+                              dmu_objset_find+0x56
+                              ztest_dmu_objset_create_destroy+0xad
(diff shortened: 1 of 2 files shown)


ahrens (approved):
can we add a test case for this?

sdb/commands/ustacks.py
```diff
@@ -0,0 +1,280 @@
...
+
+        Optionally, the command can filter stacks, displaying only
+        those that contain a given function.
+
+        The command returns all thread IDs that matched the
+        filter.
+
+    EXAMPLES
+        Print the call stacks for all threads
+    TODO
```
sdimitro: Leftover TODO

sdb/commands/ustacks.py
```diff
@@ -0,0 +1,280 @@
...
+                              thread_generic_wrapper+0x74
+                              kthread+0x121
+                              ret_from_fork+0x35
+
+
+        Count the number of stacks in the zfs module
+
+           sdb> stacks -m zfs | count
+           (unsigned long long)12
+
```
sdimitro: Per your commit description we don't implement `-m` correct?

sdb/commands/ustacks.py
```diff
@@ -0,0 +1,280 @@
...
+            0xffff9521b3f43b80 INTERRUPTIBLE         1
+                              __schedule+0x24e
+                              schedule+0x2c
+                              schedule_timeout+0x15d
+                              __cv_timedwait_common+0xdf
+                              __cv_timedwait_sig+0x16
+                              l2arc_feed_thread+0x66
+                              thread_generic_wrapper+0x74
+                              kthread+0x121
+                              ret_from_fork+0x35
```
sdimitro: Can we add some test examples here?

sdb/commands/ustacks.py
```diff
@@ -0,0 +1,280 @@
...
+            # generic "ValueError" exception, so we may wind up
+            # masking other "ValueError" exceptions that are not due
+            # to unwinding the stack of a running task.
+            #
+            # We can't check the state of the task here, and verify
+            # it's in the "R" state, since that state can change in
+            # between the point where the "ValueError" exception was
+            # originally raised, and here where we'd verify the
+            # state of the task; i.e. it could have concurrently
+            # transitioned from running to some other state.
```
sdimitro: That could be something worth mentioning the `man/help` message above; i.e. "the running thread may not come up on live processes"

BTW is that the case for core dumps too?

sdb/commands/ustacks.py
```diff
@@ -0,0 +1,280 @@
...
+                          proc_reg_write+0x43
+                          __vfs_write+0x1b
+                          vfs_write+0xb9
+                          vfs_write+0xb9
+                          ksys_write+0x67
+                          __x64_sys_write+0x1a
+                          __x64_sys_write+0x1a
+                          __x64_sys_write+0x1a
+                          do_syscall_64+0x57
+                          entry_SYSCALL_64+0x94
```
sdimitro: New example here?

sdimitro (changes requested):

sdimitro (approved):

sdimitro:
@ahrens @pcd1193182

In terms of testing we can add a userland core dump to our test suite the same way we have kernel ones but the testing infrastructure will need to be adjusted a bit more. I can slowly start working on this. In the meantime I'm good with merging this.

EDIT: Paul if you have a core dump that you used for development, you can try archiving it with `savedump` and send me over the file.

PR #331, opened 2023-07-05
DLPX-86794 sdb: zio broken after upstream ZFS commit

### LINKED ISSUE #328: AttributeError: 'zio_t' has no member 'io_parent_count'
```
sdb> zio
ADDRESS                        TYPE  STAGE            WAITER             TIME_ELAPSED
sdb encountered an internal error due to a bug. Here's the
information you need to file the bug:
----------------------------------------------------------
Target Info:
        ProgramFlags.IS_LIVE|IS_LINUX_KERNEL
        Platform(<Architecture.X86_64: 1>, <PlatformFlags.IS_LITTLE_ENDIAN|IS_64_BIT: 3>)

  File "/usr/lib/python3/dist-packages/sdb/internal/repl.py", line 107, in eval_cmd
    for obj in invoke([], input_):
  File "/usr/lib/python3/dist-packages/sdb/pipeline.py", line 152, in invoke
    yield from execute_pipeline(first_input, pipeline)
  File "/usr/lib/python3/dist-packages/sdb/pipeline.py", line 84, in execute_pipeline
    yield from massage_input_and_call(pipeline[-1], this_input)
  File "/usr/lib/python3/dist-packages/sdb/pipeline.py", line 67, in massage_input_and_call
    yield from cmd.call(objs)
  File "/usr/lib/python3/dist-packages/sdb/command.py", line 413, in call
    yield from self.__invalid_memory_objects_check(
  File "/usr/lib/python3/dist-packages/sdb/command.py", line 358, in __invalid_memory_objects_check
    for obj in objs:
  File "/usr/lib/python3/dist-packages/sdb/command.py", line 625, in _call
    self.pretty_print(self.caller(objs))
  File "/usr/lib/python3/dist-packages/sdb/commands/zfs/zio.py", line 86, in pretty_print
    for zio in objs:
  File "/usr/lib/python3/dist-packages/sdb/command.py", line 586, in caller
    yield from self.no_input()
  File "/usr/lib/python3/dist-packages/sdb/commands/zfs/zio.py", line 139, in no_input
    if zio.io_parent_count == 0:
AttributeError: 'zio_t' has no member 'io_parent_count'
----------------------------------------------------------
```

sdimitro:
Our `zio` command broke from the commit introduced by the following upstream PR: https://github.com/openzfs/zfs/pull/14948

The problem is that the PR removed `io_parent_count` from `struct zio` and we are using this field to find the top-level ZIOs (e.g. ZIOs with no parents) in the system.

This patch changes our SBD command to check whether the `io_parent_list` is populated instead.

Closes #328

### DIFF
--- a/sdb/commands/zfs/zio.py
+++ b/sdb/commands/zfs/zio.py
@@ -124,6 +124,14 @@ class Zio(sdb.Locator, sdb.PrettyPrinter):
                 yield from self.from_zio(c.zl_parent)
         self.level -= 1
 
+    @staticmethod
+    def zio_has_parents(zio: drgn.Object) -> bool:
+        parent_list = zio.io_parent_list.list_head.address_of_()
+        first_parent = parent_list.next
+        if parent_list != first_parent:
+            return True
+        return False
+
     def no_input(self) -> drgn.Object:
         if self.args.parents:
             raise sdb.CommandInvalidInputError(
@@ -136,5 +144,5 @@ class Zio(sdb.Locator, sdb.PrettyPrinter):
             [sdb.Walk(), sdb.Cast(["zio_t *"])],
         )
         for zio in zios:
-            if zio.io_parent_count == 0:
+            if not self.zio_has_parents(zio):
                 yield from self.from_zio(zio)


sdb/commands/zfs/zio.py
```diff
@@ -124,6 +124,14 @@ def from_zio(self, zio: drgn.Object) -> Iterable[drgn.Object]:
                 yield from self.from_zio(c.zl_parent)
         self.level -= 1
 
+    @staticmethod
+    def zio_has_parents(zio: drgn.Object) -> bool:
+        parent_list = zio.io_parent_list.list_head.address_of_()
+        first_parent = parent_list.next
```
grwilson, line 130: Can you just check to see if `parent_list` is not NULL?

ahrens: `io_parent_list` is not a pointer, it's a `list_t`, so it can't be NULL.

sdb/commands/zfs/zio.py
```diff
@@ -124,6 +124,14 @@ def from_zio(self, zio: drgn.Object) -> Iterable[drgn.Object]:
...
+        if parent_list != first_parent:
```
ahrens, line 131: [nit] this could be simplified to `return parent_list.next != parent_list`, I think.

ahrens (approved):

sdb/commands/zfs/zio.py
sdimitro: I tried doing that but `mypy` was complaining that it's not `bool` type :/ so I changed it to this

sdb/commands/zfs/zio.py
grwilson: I was wondering about `parent_list` which comes from `address_of_()`. I thought that returned a pointer to an Object.

ahrens: Yes, but `&zio->io_parent_list.list_head` can't be NULL.  If the zio was NULL then it would be `NULL + offsetof(zio_t, io_parent_list) + offsetof(list_t, list_head)` which is a pointer but not NULL (it's around 0x240).

mmaybee (approved):

grwilson (approved):

PR #343, opened 2023-11-10
zio command should also show any I/Os which are in the DONE stage

grwilson:


prakashsurya (approved):
do you know why we originally hit them? should we have a flag to show/hide them? just curious if we're regressing in some workflow; e.g. maybe there's "too many" in the "done" stage, making it more difficult to see the non-done IOs?

change LGTM, but I'm not sure I'm experienced enough to say if there'll be any adverse affects to this change or not.

we don't need to update any of the test cases? if not, perhaps that means we should add some tests for this command....

grwilson:
I think the assumption was that a "DONE" zio is not relevant but if it's still around it may be calling a done func and you may want to be able to track down what it's doing. I don't see any reason why you wouldn't want to show it.

prakashsurya:
cool, OK.. that makes sense to me.

mmaybee (approved):

PR #347, opened 2024-12-02
maintenance: notice & redirect

sdimitro:
For some reason the Delphix GitHub actions that include a Dockerfile stopped working from this repo so there is an action that fails again.

I don’t use `sdb` regularly anymore but I still occasionally use it for personal projects on my raspberry pi & NUC and have added a couple of commands since I left Delphix. For this reason I decided to “hard-fork” it on GitHub and open this PR that redirects people visiting this repo to my new fork which I expect would be more active.

BTW if this gets merged I’m willing to accept any kind of PR on my new fork that helps keep the Delphix infrastructure working if you end up pointing linux-pkg to it.

Let me know if there are any reasons against this in the comments.


sdimitro:
Cc @sebroy @pcd1193182 @prakashsurya @jwk404

If any of you is giving the second approval could you also merge the PR (I don’t have the permissions to do that even if the 2 shipits are there).

PR #353, opened 2025-03-12
DLPX-93639 LTS 24.04: Upgrade fails due to drgn conflict with upstream python3-drgn package

prakashsurya:
<details open>
<summary><h2> Problem </h2></summary>
Upgrade from 20.04 to 24.04 fails due to package name difference of our internal version of the "drgn" package, and ubuntu's package.
</details>

<details open>
<summary><h2> Solution </h2></summary>
The solution here, is to keep using our internal "drgn" package, rather than ubuntu's.
</details>

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,5 +7,5 @@ Build-Depends: debhelper (>= 9), dh-python, python3
 
 Package: sdb
 Architecture: any
-Depends: libkdumpfile10, python3-drgn, ${shlibs:Depends}, ${python3:Depends}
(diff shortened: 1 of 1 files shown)


palash-gandhi (approved):
Should we remove delphix-platform's dependency on drgn?

prakashsurya:
> Should we remove delphix-platform's dependency on drgn?
we could.. drgn can also be used directly, e.g. via a python script.. I'll leave it for now, but we definitely could do that..

PR #352, opened 2025-02-27
DLPX-84435 Support printing local variables

mmaybee:
<details open>
<summary><h2> Problem </h2></summary>

Need to be able to find local variables and their values
</details>

<details open>
<summary><h2> Solution </h2></summary>

This PR provodes 4 new commands: `trace`, `frame`, `locals`, and `registers`
```
sdb> help trace
SUMMARY
    trace [-h] [<task address>]
    
    Given a task_struct, trace the thread's execution state.
    
    positional arguments:
      <task address>  trace this task if no input
    
    optional arguments:
      -h, --help      show this help message and exit
    
ALIASES
    trace, bt

FIELDS
    frame# - frame number in backtrace (most recent first)
    addr - address of the frame
    function - function name in the frame (or symbol if no function)
    file - file name (if available) containing the function
    line - line number in the file

EXAMPLE
    sdb> trace 0xffff94614e796000
    TASK: 0xffff94614e796000 INTERRUPTIBLE PID: 268
    #0  0xffffffffa0c549e8 in context_switch() at core.c:5038:2 (inlined)
    #1  0xffffffffa0c549e8 in __schedule() at core.c:6384:8
    #2  0xffffffffa0c54fe9 in schedule() at core.c:6467:3
    #3  0xffffffffa0c595c2 in schedule_timeout() at timer.c:2116:2
    #4  0xffffffffc04836ae in __cv_timedwait_common() at spl-condvar.c:250:15
    #5  0xffffffffc0483829 in __cv_timedwait_idle() at spl-condvar.c:307:7
    #6  0xffffffffc0571554 in l2arc_feed_thread() at arc.c:9460:10
    #7  0xffffffffc048c8d1 in thread_generic_wrapper() at spl-thread.c:61:3
    #8  0xffffffffa00efb17 in kthread() at kthread.c:334:9
    #9  0xffffffffa0004c3f (ret_from_fork+0x1f/0x2d) at entry_64.S:287

sdb> help frame
SUMMARY
    frame [-h] [-v] [<frame>]
    
    Given a task_struct and frame, return requested stack frame
    
    positional arguments:
      <frame>        set local context to this frame number
    
    optional arguments:
      -h, --help     show this help message and exit
      -v, --verbose  print full file path for frame
    
ALIASES
    stackframe, frame, f

EXAMPLE
    sdb> frame 7
    #7  0xffffffffc048c8d1 in thread_generic_wrapper (arg=0xffff94614f3cc400) at spl-thread.c:61:3

[...]

PR #356, opened 2025-11-18
DLPX-95965 _framestr() parsing is fragile - should use drgn StackFrame API properties

dbjwhs-perforce:
### Problem

@mmaybee called out in https://github.com/delphix/sdb/pull/352 that the `_framestr()` function was using string parsing on `str(frame)` output to extract frame information. This approach was brittle and could break if drgn's string representation format changed.

### Solution

This PR replaces the string parsing with direct use of drgn StackFrame API properties:

- `frame.pc` - Program counter (address)
- `frame.function_name` - Function name (or None if unavailable)
- `frame.symbol()` - Symbol object for fallback name lookup
- `frame.source()` - Returns `(filename, line, column)` tuple
- `frame.is_inline` - Boolean for inline frame detection

### Changes

1. **Updated `_framestr()` signature**: Added required `frame_index: int` parameter.

2. **Rewrote frame formatting logic**: Uses API properties directly with error handling via try/except for `LookupError`

3. **Updated call sites**: 
   - Changed `for frame in stack_trace:` to `for frame_index, frame in enumerate(stack_trace):`
   - Updated calls from `_framestr(frame, False)` to `_framestr(frame, False, frame_index)`
   - Updated calls from `_framestr(frame, True)` to `_framestr(frame, True, self.frame_id)`

### Testing

Tested with crash dump from kernel 6.14.0-27-dx2025082915-8ba235c2d-generic:

```
sdb> trace 0xffff8f1a285e5400
TASK: 0xffff8f1a285e5400 INTERRUPTIBLE PID: 12546
#0  0xffffffff92718fd0 in context_switch() at core.c:5378:2 (inlined)
#1  0xffffffff92718fd0 in __schedule() at core.c:6765:8
#2  0xffffffff92719389 in __schedule_loop() at core.c:6842:3 (inlined)
#3  0xffffffff92719389 in schedule() at core.c:6857:2
#4  0xffffffffc06c8a40
#5  0xffffffff917ab35b in kthread() at kthread.c:464:9
#6  0xffffffff916d9dd4 in ret_from_fork() at process.c:153:3
```

All features working correctly:
- Frame indices from enumerate
- Addresses from `frame.pc`
- Function names from `frame.function_name`
- Source locations from `frame.source()` showing `file:line:col`
- Inline markers from `frame.is_inline`

### Benefits

- **More robust**: No longer dependent on string format
- **More maintainable**: Uses documented API instead of parsing heuristics
- **Better error handling**: Explicit try/except for missing data
- **Clearer code**: Intent is obvious from property names

### Verified Command Compatibility

All related stack inspection commands continue to work correctly with the updated `_framestr()` implementation:

**`trace` command** - Displays full stack trace with frame indices, addresses, functions, and source locations:
```
sdb> trace 0xffff8f1a285e5400
TASK: 0xffff8f1a285e5400 INTERRUPTIBLE PID: 12546
#0  0xffffffff92718fd0 in context_switch() at core.c:5378:2 (inlined)
#1  0xffffffff92718fd0 in __schedule() at core.c:6765:8
#2  0xffffffff92719389 in __schedule_loop() at core.c:6842:3 (inlined)
#3  0xffffffff92719389 in schedule() at core.c:6857:2
...
```

**`frame` command** - Shows detailed frame information with function arguments (uses `_framestr()` with `args=True`):
```
sdb> frame 0
#0  0xffffffff92718fd0 in context_switch (rq=0xffff8f566bbb6e80, prev=0xffff8f1a285e5400, 
    next=0xffff8f1840f32a00, rf=0xffffa36c7599bd20) at core.c:5378:2 (inlined)
```

**`locals` command** - Displays local variables in the current frame:
```
sdb> locals
rf = (struct rq_flags *)0xffffa36c7599bd20
next = (struct task_struct *)0xffff8f1840f32a00
prev = (struct task_struct *)0xffff8f1a285e5400
rq = (struct rq *)0xffff8f566bbb6e80
```

**`registers` command** - Shows CPU register values at the current frame:
```
sdb> registers
rbx = 18446620200070246016
rbp = 18446642284957646200
rsp = 18446642284957646096
r12 = 18446619941242033152
r13 = 18446619933064505856
r14 = 0
r15 = 18446619941242036240
rip = 18446744071871500240
```

These commands demonstrate that the refactored `_framestr()` function maintains full backward compatibility while providing more robust frame formatting through the drgn StackFrame API.

### DIFF
--- a/sdb/commands/linux/threads.py
+++ b/sdb/commands/linux/threads.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2020 Delphix
+# Copyright 2020, 2025 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -110,42 +110,57 @@ class KernelThreads(sdb.Locator, sdb.PrettyPrinter):
         yield from for_each_task(sdb.get_prog())
 
 
-def _framestr(frame: drgn.Object, args: bool) -> str:
-    # This parsing is a bit fragile. It depends on the string
-    # for the frame to have the following format:
-    # <id> in <addr> <symbol> [in <function>] at <filepath> [(inlined)]
-    # It pulls off the last element of the file path and returns:
-    # <id> <addr> in <function>() at <filename> [(inlined)]
-    # or (if no function is available):
-    # <id> <addr> <symbol> at <filename> [(inlined)]
-    inlined = ""
-    parts = str(frame).split()
-    id = parts[0]
-    if len(id) < 3:
-        id = id + " "
-    addr = parts[2]
-    # Just have an address
-    if len(parts) < 4:
-        return id + " " + addr
-    # No function or file path available? Just return the symbol
-    if len(parts) < 5:
-        return id + " " + addr + " " + parts[3]
-    if parts[4] != "in":
-        addr = addr + " " + parts[3]
-        function = ""
-        filepath = parts[5]
+def _framestr(frame_index: int, frame: drgn.StackFrame, args: bool) -> str:
+    """
+    Format a stack frame using drgn StackFrame API properties.
+
+    Args:
+        frame_index: Frame index number
+        frame: drgn StackFrame object
+        args: If True, show function arguments
+
+    Returns:
+        Formatted frame string like: "#0  0xaddr in function() at file.c:line:col (inlined)"
+    """
+    # Format frame index with padding for alignment
+    id_str = f"#{frame_index}"
+    if len(id_str) < 3:
+        id_str = id_str + " "
+
+    # Get program counter (address) - pc is always available as a Final[int] attribute
+    addr = hex(frame.pc)
+
+    # Get function/symbol name
+    function_name = frame.function_name
+    if function_name is None:
+        try:
+            function_name = frame.symbol().name
+        except LookupError:
+            # No symbol available, just return address
+            return f"{id_str} {addr}"
+
+    # Build function string
+    if args:
+        function_str = " in " + _funcstr(frame)
     else:
-        function = " in " + parts[5] + "()"
-        if args:
-            function = " in " + _funcstr(frame)
-        filepath = parts[7]
-    filename = " at " + filepath.split("/")[-1]
-    if frame.is_inline:
-        inlined = " (inlined)"
-    return id + " " + addr + function + filename + inlined
+        function_str = f" in {function_name}()"
 
+    # Get source location
+    try:
+        filename, line, column = frame.source()
+        # Extract just the filename from the full path
+        filename_only = filename.split("/")[-1] if filename else "??"
+        location_str = f" at {filename_only}:{line}:{column}"
+    except LookupError:
+        # No source information available
+        location_str = ""
 
-def _funcstr(frame: drgn.Object) -> str:
+    # Add inline marker if applicable
+    inline_str = " (inlined)" if frame.is_inline else ""
+
+    return id_str + " " + addr + function_str + location_str + inline_str
+
+def _funcstr(frame: drgn.StackFrame) -> str:
     name = frame.name
     if name is None:
         try:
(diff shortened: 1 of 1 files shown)


sdb/commands/linux/threads.py
```diff
@@ -110,40 +110,59 @@ def no_input(self) -> Iterable[drgn.Object]:
...
-    if len(parts) < 4:
-        return id + " " + addr
-    # No function or file path available? Just return the symbol
-    if len(parts) < 5:
-        return id + " " + addr + " " + parts[3]
-    if parts[4] != "in":
-        addr = addr + " " + parts[3]
-        function = ""
-        filepath = parts[5]
+def _framestr(frame: drgn.Object, args: bool, frame_index: int) -> str:
```
mmaybee: nit: consider rearranging the arguments: (frame_index, frame, args)

sdb/commands/linux/threads.py
```diff
@@ -110,40 +110,59 @@ def no_input(self) -> Iterable[drgn.Object]:
...
+    # Format frame index with padding for alignment
+    id_str = f"#{frame_index}"
+    if len(id_str) < 3:
+        id_str = id_str + " "
+
+    # Get program counter (address)
+    try:
+        addr = hex(frame.pc)
+    except LookupError:
+        # No PC available
```
mmaybee: Should we return some sort of indicator that the pc is missing (this is pretty unexpected)? Perhaps something like:
```
    id_str = id_str + "<no address found>"
```

mmaybee (approved):

sdb/commands/linux/threads.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2020 Delphix
+# Copyright 2025 Delphix
```
lyriclake: This can be 
```
# Copyright 2020, 2025 Delphix
```
to match what we've done elsewhere, e.g.,https://github.com/delphix/sdb/blob/6306f9fd1298e3260216eb6f2b36fcff3e8ba9e0/sdb/commands/linux/slabs.py#L2

sdb/commands/linux/threads.py
```diff
@@ -110,40 +110,59 @@ def no_input(self) -> Iterable[drgn.Object]:
...
-    if len(parts) < 4:
-        return id + " " + addr
-    # No function or file path available? Just return the symbol
-    if len(parts) < 5:
-        return id + " " + addr + " " + parts[3]
-    if parts[4] != "in":
-        addr = addr + " " + parts[3]
-        function = ""
-        filepath = parts[5]
+def _framestr(frame_index: int, frame: drgn.Object, args: bool) -> str:
```
lyriclake: Can the type hint be `frame: drgn.StackFrame` here?

sdb/commands/linux/threads.py
```diff
@@ -110,40 +110,59 @@ def no_input(self) -> Iterable[drgn.Object]:
...
+    """
+    # Format frame index with padding for alignment
+    id_str = f"#{frame_index}"
+    if len(id_str) < 3:
+        id_str = id_str + " "
+
+    # Get program counter (address)
+    try:
+        addr = hex(frame.pc)
+    except LookupError:
```
lyriclake: Can this actually happen in the case of `pc`? Looking at https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.StackFrame `pc` is defined to be a `Final[int]` attribute, vs the others like `source()` which are getter functions that declare `Raises: LookupError`.

I suppose hypothetically the object could be missing the `pc` attribute entirely, but then I'd expect that doing `frame.pc` would result in `AttributeError` -- though I don't know whether drgn is ever expected to create a `StackFrame` object like that.

**edit:** just as a side note on this point, I noticed further down in the file that in the `KernelTrace` case we do 
```
                if frame.pc == 0:
                    # Note: We filter out frames with zero program counter,
                    # as we often see the stack trace padded with zeros
                    continue
```
right before the call to `_framestr()`. So, if it _were_ possible for `frame.pc` to fail, it would have failed already before entering this function.

lyriclake (approved):
Thanks for taking this on!

sdb/commands/linux/threads.py
dbjwhs-perforce: done

sdb/commands/linux/threads.py
dbjwhs-perforce: much better thanks.

sdb/commands/linux/threads.py
dbjwhs-perforce: dope. this is why I love code reviews, thank you @lyriclake you are correct, fixed!

dbjwhs-perforce:
@mmaybee if you look back at this heads up I removed the `<no address found>` handling because @lyriclake pointed out that frame.pc is a `Final[int]` attribute that cannot fail, and we already access it without protection at line 268. So all that got removed.

PR #357, opened 2025-11-18
DLPX-95958 sdb stacks command doesn't always work

dbjwhs-perforce:
### Problem

The `stacks -m <module>` command failed on kernel 6.14.0-27 with:
```
AttributeError: 'struct module' has no member 'core_layout'
```

This prevented filtering stack traces by kernel module name, a critical debugging feature for isolating issues in specific modules like ZFS.

### Root Cause

Kernel 6.14+ restructured `struct module` memory layout, changing from a single `core_layout` field to a `mem[]` array of memory segments. The `find_module_memory_segment()` function directly accessed `mod.core_layout.base` and `mod.core_layout.size`, which no longer exist in newer kernels.

### Solution

Refactored `find_module_memory_segment()` in `sdb/commands/linux/stacks.py` to use **drgn's Module API** instead of directly examining kernel structures:

```python
mod = sdb.get_prog().module(mod_name)
ranges = mod.address_ranges
```

The `Module.address_ranges` property returns a tuple of (start, end) address pairs for all memory segments, abstracting away kernel version differences. This eliminates the need for version detection or manual structure inspection.

### Testing

Tested with crash dump from kernel 6.14.0-27-dx2025082915-8ba235c2d-generic:

**Test 1 - Non-existent module:**
```
sdb> stacks -m nonexistent
sdb: stacks: module 'nonexistent' doesn't exist or isn't currently loaded
```
Result: Correctly reports missing module

**Test 2 - ZFS module (exists):**
```
sdb> stacks -m zfs -t INTERRUPTIBLE
0xffff8f1855432a00 INTERRUPTIBLE       355
                  context_switch (inlined)
                  __schedule+0x2c0
                  __schedule_loop (inlined)
                  schedule+0x29
                  0xffffffffc06c8a40+0x0    <- ZFS module address
                  kthread+0xfb
                  ret_from_fork+0x44
                  ret_from_fork_asm+0x1a
```
Result: Successfully finds threads with ZFS stack frames

**Module address ranges verification:**
Using `prog.module("zfs").address_ranges`:
```
Range 0: start=0xffffffffc06fd000, end=0xffffffffc0a13000, size=3235840
Range 1: start=0xffffffffc0a14000, end=0xffffffffc0c78000, size=2506752
Range 2: start=0xffffffffc0c79000, end=0xffffffffc0d6c000, size=995328
Range 3: start=0xffffffffc062f000, end=0xffffffffc0630000, size=4096

Overall range: 0xffffffffc062f000 to 0xffffffffc0d6c000 (size=7589888)
```
Result: Correctly calculates memory range across all segments

### DIFF
--- a/sdb/commands/linux/stacks.py
+++ b/sdb/commands/linux/stacks.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2019 Delphix
+# Copyright 2025 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -21,7 +21,6 @@ from typing import Dict, Iterable, List, Optional, Tuple
 from collections import defaultdict
 
 import drgn
-from drgn.helpers.linux.list import list_for_each_entry
 from drgn.helpers.linux.pid import for_each_task
 from drgn.helpers.linux.sched import task_state_to_char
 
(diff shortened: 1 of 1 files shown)


mmaybee:
Would it be be possible to use the drgn API for the module class: `address_ranges` instead of trying to examine contents of the system structure (and worry about versioning, etc)?

dbjwhs-perforce:
> Would it be be possible to use the drgn API for the module class: `address_ranges` instead of trying to examine contents of the system structure (and worry about versioning, etc)?

Yes, confirmed; This will be better, will get this out later today, thanks.

dbjwhs-perforce:
@mmaybee V2 in, much better thanks.

mmaybee (approved):

tonynguien (approved):

